PR #20, opened 2023-12-07
correct product name

lnxbil:
The product has to adhere to the Proxmox Band Guideline, so Proxmox VE is the product name. Maybe it is present on other pages too. Proxmox itself is the company and does not reference to concrete product.

 I don't know if the name of the repo has to be changed too, yet I suggest it.


wikkyk:
Thanks for bringing this up. I will close this PR in favour of #21 which is more complete.

PR #28, opened 2023-12-14
Fixes pause of IPAddressClaim

mcbenjemaa:
## Overview 

IPAddressClaim is not paused due to the pausing implementation was not correct:

The correct check is the clusterName label:

https://github.com/kubernetes-sigs/cluster-api-ipam-provider-in-cluster/blob/b8ccdf8c362cd51ee120e50f8ec2b637f02e3755/internal/controllers/ipaddressclaim.go#L180

### DIFF
--- a/pkg/kubernetes/ipam/ipam.go
+++ b/pkg/kubernetes/ipam/ipam.go
@@ -152,7 +152,7 @@ func (h *Helper) GetGlobalInClusterIPPool(ctx context.Context, ref *corev1.Typed
 }
 
 // CreateIPAddressClaim creates an IPAddressClaim for a given object.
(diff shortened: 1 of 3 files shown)


pkg/kubernetes/ipam/ipam.go
```diff
@@ -200,15 +200,15 @@ func (h *Helper) CreateIPAddressClaim(ctx context.Context, owner client.Object,
 
 	// Ensures that the claim has a reference to the cluster of the VM to
 	// support pausing reconciliation.
-	annotations := map[string]string{
-		clusterv1.ClusterNameAnnotation: h.cluster.GetName(),
+	labels := map[string]string{
+		clusterv1.ClusterNameLabel: h.cluster.GetName(),
```
avorima, line 204: This only works if the ProxmoxCluster has the same name as the CAPI cluster

avorima (approved):

PR #30, opened 2023-12-14
ADD Advanced setups Guide

mcbenjemaa:
This PR adds:

* Docs for Advanced setups.


templates/cluster-template.yaml
```diff
@@ -123,6 +123,17 @@ spec:
       templateID: ${TEMPLATE_VMID}
       format: "qcow2"
       full: true
+      numSockets: ${NUM_SOCKETS:=2}
```
mcbenjemaa, line 126: I removed the xl-nodes template cause our usage docs mention the customization of the machine while not supported by the default template.

wikkyk (approved):

PR #32, opened 2023-12-14
update troubleshooting + quickstart docs to aid with capi incompatibilities

pborn-ionos:


docs/Usage.md
```diff
@@ -101,7 +101,7 @@ the `EXP_CLUSTER_RESOURCE_SET` is required if you want to deploy CNI using clust
 
 Once you have access to a management cluster, you can initialize Cluster API with the following:
 ```
-clusterctl init --infrastructure proxmox --ipam in-cluster
+clusterctl init --infrastructure proxmox --ipam in-cluster --core cluster-api:v1.5.3
```
mcbenjemaa, line 104: @avorima 
Do you agree with pinning the capi version in the quick-start docs?
while our compatibility versions matrix mentions that v1.6 not yet supported?
https://github.com/ionos-cloud/cluster-api-provider-proxmox/tree/main?tab=readme-ov-file#compatibility-with-cluster-api-and-kubernetes-versions


In my point of view we should plan the support for v1.6 very soon.

mcbenjemaa: Let's wait for @avorima feedback.

avorima: Yes, since those commands are meant to be copied into a terminal.

wikkyk: ok then

wikkyk (approved):

PR #38, opened 2023-12-16
fix: prevent deletion attempts for non-existent vmID -1

### LINKED ISSUE #31: ProxmoxMachine deletion fails and loops with `cannot find vm with id -1`
CAPMOX can get stuck during `ProxmoxMachine` deletion, when a VM has never been provisioned. It then tries to delete a VM with the ID `-1`, which does not (and probably can not) exist.

`
capmox-controller-manager-6c57697857-6lxsg manager E1214 12:10:49.116477       1 controller.go:324] "Reconciler error" err="cannot find vm with id -1: bad request: 400 Parameter verification failed. - {\"vmid\":\"invalid format - value does not look like a valid VM ID\\n\"}" controller="proxmoxmachine" controllerGroup="infrastructure.cluster.x-k8s.io" controllerKind="ProxmoxMachine" ProxmoxMachine="capmox/test-worker-kf5pm" namespace="capmox" name="test-worker-kf5pm" reconcileID="f1d47249-fe64-49ba-9a3f-9d69b8e8fbce"
`

If a VMID for a VM is not known, instead of computing it to be `-1` and attempting to delete it (which will not work, as you see in the error message above), just accept the deletion request and complete it.

**Environment:**

- Cluster-api-provider-proxmox version: v0.1.0
- Kubernetes version: (use `kubectl version`): v1.28.4
- OS (e.g. from `/etc/os-release`):

pborn-ionos:
This should fix #31. It simply checks if the passed vmID is `< 100` (which is not allowed by Proxmox) and then returns the `does not exist` error, which we'd expect anyways, in case a VM does not exist. That way it also spares us from making unnecessary API calls to Proxmox.

Alternative solution would be to just reword our  error in the same function from `cannot find vm with id %d` to `vm with id %d does not exist`, which would effectively have the same functional effect in our controller, but would in turn make unnecessary API calls to Proxmox for a VM that could've never existed.

Both solutions would work with https://github.com/ionos-cloud/cluster-api-provider-proxmox/blob/5f6d700a2d39e0c54818354485c8fbadf8cb442d/internal/service/vmservice/delete.go#L52C1-L55C2

### DIFF
--- a/pkg/proxmox/goproxmox/api_client.go
+++ b/pkg/proxmox/goproxmox/api_client.go
@@ -138,6 +138,12 @@ func (c *APIClient) FindVMResource(ctx context.Context, vmID uint64) (*proxmox.C
 
 // DeleteVM deletes a VM based on the nodeName and vmID.
 func (c *APIClient) DeleteVM(ctx context.Context, nodeName string, vmID int64) (*proxmox.Task, error) {
+	// A vmID can not be lower than 100.
+	// If the provided vmID is lower (like -1 in issue #31), just error out without calling the API.
+	if vmID < 100 {
+		return nil, fmt.Errorf("vm with id %d does not exist", vmID)
+	}
+
 	node, err := c.Node(ctx, nodeName)
 	if err != nil {
 		return nil, fmt.Errorf("cannot find node with name %s: %w", nodeName, err)


mcbenjemaa (approved):
LGTM,

That's correct.
The GetVirtualMachine ID returns -1 if the vm is not yet created. 

We could also put this check in the reconciler but also here, is okay.

PR #42, opened 2023-12-17
feat: implement setting mtu for virtio network device

lucasl0st:
For some scenarios it is required to set the MTU of the proxmox VM network interface, especially when running behind a VPN.
This adds a `MTU` field to the `NetworkDevice` struct of the `ProxmoxMachine` CRD.

See https://git.proxmox.com/?p=qemu-server.git;a=commit;h=61a14cde8d568e552d3deaab2da76b479b8aca7b

Sets the MTU value of a network machine network device:
image


internal/service/vmservice/vm.go
```diff
@@ -191,16 +191,20 @@ func reconcileVirtualMachineConfig(ctx context.Context, machineScope *scope.Mach
 	if machineScope.ProxmoxMachine.Spec.Network != nil && shouldUpdateNetworkDevices(machineScope) {
```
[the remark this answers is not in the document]
mcbenjemaa: please also carefully change `extractNetworkModelAndBridge()` so it returns the right thing when MTU is not set.
maybe also it can return a struct instead.

api/v1alpha1/proxmoxmachine_types.go
```diff
@@ -216,6 +216,13 @@ type NetworkDevice struct {
 	// +kubebuilder:validation:Enum=e1000;virtio;rtl8139;vmxnet3
 	// +kubebuilder:default=virtio
 	Model *string `json:"model,omitempty"`
+
+	// MTU is the network device Maximum Transmission Unit.
+	// Only works with virtio Model.
+	// +optional
+	// +kubebuilder:validation:Minimum=1
+	// +kubebuilder:default=1500
+	MTU *uint16 `json:"mtu,omitempty"`
```
mcbenjemaa, line 225: Can you add a test case in the `proxmoxmachine_types_test.go`
something like: 
`Should not allow machine with network device mtu less than 1`

api/v1alpha1/proxmoxmachine_types.go
```diff
@@ -216,6 +216,13 @@ type NetworkDevice struct {
 	// +kubebuilder:validation:Enum=e1000;virtio;rtl8139;vmxnet3
 	// +kubebuilder:default=virtio
 	Model *string `json:"model,omitempty"`
+
+	// MTU is the network device Maximum Transmission Unit.
+	// Only works with virtio Model.
+	// +optional
+	// +kubebuilder:validation:Minimum=1
```
mcbenjemaa, line 223: minimum=1 is okay,
but you can set value like 100

we probably have to check this in the webhook.

mcbenjemaa: `Should not allow mtu less than 1000 and more than 65520`

mcbenjemaa (changes requested):

api/v1alpha1/proxmoxmachine_types.go
lucasl0st: I think less than 1000 should be allowed because 1 equals the MTU of the network bridge: 

https://git.proxmox.com/?p=qemu-server.git;a=commit;h=61a14cde8d568e552d3deaab2da76b479b8aca7b
`Special value: 1 = set the same value than the bridge`

lucasl0st: Ah okay now I get it, I will try to implement that. Thanks for the review!

mcbenjemaa: Thank you for addressing this. 💯

internal/service/vmservice/utils_test.go
```diff
@@ -54,42 +54,106 @@ func TestExtractUUID(t *testing.T) {
...
+func TestExtractNetworkMTU(t *testing.T) {
+	type match struct {
+		test     string
+		expected uint16
+	}
+
+	goodstrings := []match{
+		{"virtio=A6:23:64:4D:84:CB,bridge=vmbr1,mtu=1500", 1500},
+		{"foo,virtio=A6:23:64:4D:84:CB,bridge=vmbr1,mtu=1500", 1500},
+		{"foo=bar,virtio=A6:23:64:4D:84:CB,bridge=vmbr1,mtu=1500", 1500},
```
mcbenjemaa, line 134: just include a test case without mtu specified

internal/service/vmservice/utils.go
```diff
@@ -78,8 +117,14 @@ func shouldUpdateNetworkDevices(machineScope *scope.MachineScope) bool {
...
-		model, bridge := extractNetworkModelAndBridge(net0)
-		if model != *machineScope.ProxmoxMachine.Spec.Network.Default.Model || bridge != machineScope.ProxmoxMachine.Spec.Network.Default.Bridge {
+
+		desiredDefault := *machineScope.ProxmoxMachine.Spec.Network.Default
+
+		model := extractNetworkModel(net0)
+		bridge := extractNetworkBridge(net0)
+		mtu := extractNetworkMTU(net0)
+
+		if model != *desiredDefault.Model || bridge != desiredDefault.Bridge || mtu != *desiredDefault.MTU {
```
mcbenjemaa, line 127: Do we need to check if mtu was specified first?

internal/service/vmservice/utils.go
```diff
@@ -91,9 +136,13 @@ func shouldUpdateNetworkDevices(machineScope *scope.MachineScope) bool {
...
 		}
-		model, bridge := extractNetworkModelAndBridge(net)
+
+		model := extractNetworkModel(net)
+		bridge := extractNetworkBridge(net)
+		mtu := extractNetworkMTU(net)
+
 		// current is different from the desired spec.
-		if model != *v.Model || bridge != v.Bridge {
+		if model != *v.Model || bridge != v.Bridge || mtu != *v.MTU {
```
mcbenjemaa, line 145: Same?

internal/service/vmservice/utils.go
lucasl0st: I dont think so because it is said to `default=1500` in the CRD (similar how the Model is set virtio as default)

internal/service/vmservice/utils.go
mcbenjemaa: I have tested it.
This will set mtu to 1500 as it's the default value even when mtu is not defined in the spec.
probably separate it from the condition.

internal/service/vmservice/utils.go
mcbenjemaa: https://github.com/ionos-cloud/cluster-api-provider-proxmox/pull/42#discussion_r1431642877

internal/service/vmservice/utils.go
```diff
@@ -55,14 +57,51 @@ func IPAddressWithPrefix(ip string, prefix int) string {
 	return fmt.Sprintf("%s/%d", ip, prefix)
 }
 
-// extractNetworkModelAndBridge returns the model & bridge out of net device input e.g. virtio=A6:23:64:4D:84:CB,bridge=vmbr1.
-func extractNetworkModelAndBridge(input string) (string, string) {
-	re := regexp.MustCompile(`([^=,]+)=([^,]+),bridge=([^,]+)`)
+// extractNetworkModel returns the model out of net device input e.g. virtio=A6:23:64:4D:84:CB,bridge=vmbr1,mtu=1500.
+func extractNetworkModel(input string) string {
```
mcbenjemaa, line 61: Suggest

```go
func extractNetworkModel(input string) string {
	re := regexp.MustCompile(`([^=,]+)=([^,]+),bridge=([^,]+)`)
	matches := re.FindStringSubmatch(input)
        if len(matches) > 1 {
		return matches[1]
	}
	return "virtio"
}
```

internal/service/vmservice/utils_test.go
```diff
@@ -54,42 +54,109 @@ func TestExtractUUID(t *testing.T) {
...
+		expected string
 	}
 
 	goodstrings := []match{
-		{"virtio=A6:23:64:4D:84:CB,bridge=vmbr1", "virtio", "vmbr1"},
-		{"foo,virtio=A6:23:64:4D:84:CB,bridge=vmbr1", "virtio", "vmbr1"},
-		{"foo=bar,virtio=A6:23:64:4D:84:CB,bridge=vmbr1", "virtio", "vmbr1"},
-		{"foo=bar,virtio=A6:23:64:4D:84:CB,bridge=vmbr1,foo", "virtio", "vmbr1"},
-		{"foo=bar,virtio=A6:23:64:4D:84:CB,bridge=vmbr1,foo=bar", "virtio", "vmbr1"},
+		{"virtio=A6:23:64:4D:84:CB,bridge=vmbr1,mtu=1500", "virtio"},
```
mcbenjemaa, line 64: test case `{"virtio=,bridge=vmbr1", "virtio"},` should be okay

internal/service/vmservice/utils.go
mcbenjemaa: Update:

```
func extractNetworkModel(input string) string {
	re := regexp.MustCompile(`([^,=]+)(?:=[^,]*)?,bridge=([^,]+)`)
	matches := re.FindStringSubmatch(input)
	if len(matches) >= 2 {
		return matches[1]
	}
	return ""
}
```

lucasl0st: why have the bridge part of this regex? the bridge is not part of the network model

mcbenjemaa: Yeah, but since the wording bridge is always in the net-config.
So, using it in the regex.
it will return the first match. so the model.

mcbenjemaa:
https://goplay.tools/snippet/5aCh1GVWyXM

lucasl0st:
As requested I removed the default MTU and applied the other suggestions, it should have the original behavior now if no MTU is set

mcbenjemaa (approved):
LGTM

thanks for your contribution

PR #43, opened 2023-12-18
ci: run sonarcloud via pull_request_target

pborn-ionos:
`on.pull_request` executes the workflow out of a forked repository. This means the `SONAR_TOKEN` secret will not be available and therefore the workflow will fail. By changing it to `on.pull_request_target`, the workflow definition (and code) from the trusted original repository is run and the secret will be available.

See https://github.blog/2020-08-03-github-actions-improvements-for-fork-and-pull-request-workflows/

This should fix issues with failed SonarCloud runs on external PRs like in https://github.com/ionos-cloud/cluster-api-provider-proxmox/actions/runs/7240484604/job/19723537444

### DIFF
--- a/.github/workflows/code-qa-sonarcloud.yaml
+++ b/.github/workflows/code-qa-sonarcloud.yaml
@@ -1,0 +1,44 @@
+name: Code QA - SonarCloud
+
+on:
+  push:
+    branches:
+      - main
+  pull_request_target:
+    types:
+      - opened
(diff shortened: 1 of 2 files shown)


mcbenjemaa:
not sure about this

mcbenjemaa:
even sonar job is not running in this PR

mcbenjemaa (approved):
LGTM

PR #46, opened 2023-12-19
docs: document token creation

pborn-ionos:
Might be a good idea to guide the enduser to create and use a less-privileged account :)


docs/Usage.md
```diff
@@ -61,6 +61,17 @@ providers:
 Before you can create a cluster, you need to configure your management cluster. 
 This is done by setting up the environment variables for CAPMOX and generating a cluster manifest.
 
+---
+**_NOTE_**: A dedicated Proxmox VE user + API token can be created by executing
```
mcbenjemaa, line 65: you can also create on the UI

mcbenjemaa (approved):

PR #50, opened 2023-12-27
feat(scheduler): support optional cpu constraints and make sort by guest count optional

tamcore:
This optionally allows the scheduler to take CPU cores into account, as well as disabling the sorting by guest count. I think with this it should be relatively simple, to, for example, include a host's actual cpu load as well.


mcbenjemaa:
Thanks for your PR, We will check this.

PR #53, opened 2024-01-08
Implement DHCP for machine network

mcbenjemaa:
THis PR adds:

- Support DHCP


api/v1alpha1/helpers.go
```diff
@@ -0,0 +1,157 @@
+/*
+Copyright 2023 IONOS Cloud.
```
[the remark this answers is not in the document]
wikkyk: wondering if the license action will complain :)

api/v1alpha1/proxmoxcluster_types.go
```diff
@@ -50,25 +48,49 @@ type ProxmoxClusterSpec struct {
...
 	// DNSServers contains information about nameservers used by machines network-config.
 	// +kubebuilder:validation:MinItems=1
 	DNSServers []string `json:"dnsServers"`
 }
 
+// IPConfig contains information about available IP config.
+type IPConfig struct {
+	// Addresses is a list of IP addresses that can be assigned. This set of
+	// addresses can be non-contiguous.
+	// +kubebuilder:validation:MinItems=1
```
avorima, line 77: Shouldn't DHCP and addresses be mutually exclusive?

api/v1alpha1/proxmoxcluster_types.go
```diff
@@ -50,25 +48,49 @@ type ProxmoxClusterSpec struct {
...
+	// +optional
+	Prefix int `json:"prefix,omitempty"`
+
+	// Gateway
+	// +optional
+	Gateway string `json:"gateway,omitempty"`
+
+	// DHCP indicates if DHCP should be used to assign IP addresses.
+	// +optional
+	DHCP bool `json:"dhcp,omitempty"`
```
avorima, line 91: This needs to be a pointer if `false` is supposed to be a valid value

avorima (changes requested):
First batch of comments. Didn't look at vmservice yet

api/v1alpha1/helpers.go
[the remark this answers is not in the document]
lubedacht: Maybe use https://github.com/denis-tingaikin/go-header in [golangci-lint](https://golangci-lint.run/usage/linters/#goheader).

You can define a year range there, and it would also make the action obsolete.

api/v1alpha1/proxmoxcluster_types.go
65278: DHCP and static IP addresses on teh same interface aren't mutually exclusive, and they also aren't in netplan.

avorima: Ok, let me rephrase. Should addresses really be required when using DHCP is an option?

api/v1alpha1/helpers.go
```diff
@@ -0,0 +1,157 @@
...
+	if !c.HasMachine(loc.Machine.Name, isControlPlane) {
+		c.addNodeLocation(loc, isControlPlane)
+	}
+}
+
+// RemoveNodeLocation removes a node location from the status.
+func (c *ProxmoxCluster) RemoveNodeLocation(machineName string, isControlPlane bool) {
+	nodeLocations := c.Status.NodeLocations
+
+	if nodeLocations == nil {
```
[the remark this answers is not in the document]
mcbenjemaa: but, this is used in `DeleteVM`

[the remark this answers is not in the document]
mcbenjemaa: right, but that can be removed

api/v1alpha1/proxmoxcluster_types.go
```diff
@@ -50,25 +48,49 @@ type ProxmoxClusterSpec struct {
...
 
+// IPConfig contains information about available IP config.
+type IPConfig struct {
+	// Addresses is a list of IP addresses that can be assigned. This set of
+	// addresses can be non-contiguous.
+	// +kubebuilder:validation:MinItems=1
+	Addresses []string `json:"addresses,omitempty"`
+
+	// Prefix is the network prefix to use.
+	// +kubebuilder:validation:Maximum=128
```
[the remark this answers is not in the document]
mcbenjemaa: I don't understand 😄 

I didn't change anything here!

the webhook already check this

api/v1alpha1/proxmoxcluster_types.go
mcbenjemaa: No, it's mutually exclusive.

api/v1alpha1/helpers.go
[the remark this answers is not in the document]
wikkyk: I've deleted the action for now. We can evaluate go-header and if we like it, I'll add the updated licence action back in.

But yes this should say 2024.

api/v1alpha1/proxmoxmachine_types.go
```diff
@@ -223,12 +223,22 @@ type NetworkDevice struct {
...
+	// +optional
+	DHCP6 bool `json:"dhcp6,omitempty"`
 }
 
 // AdditionalNetworkDevice the definition of a Proxmox network device.
-// +kubebuilder:validation:XValidation:rule="self.ipv4PoolRef != null || self.ipv6PoolRef != null",message="at least one pool reference must be set, either ipv4PoolRef or ipv6PoolRef"
+// +kubebuilder:validation:XValidation:rule="(self.ipv4PoolRef != null || self.ipv6PoolRef != null || self.dhcp4 || self.dhcp6)",message="at least dhcp and/or one pool reference must be set, either ipv4PoolRef or ipv6PoolRef"
 type AdditionalNetworkDevice struct {
-	NetworkDevice `json:",inline"`
+	*NetworkDevice `json:",inline"`
```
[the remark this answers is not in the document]
mcbenjemaa: right

internal/controller/proxmoxcluster_controller.go
```diff
@@ -243,3 +244,16 @@ func (r *ProxmoxClusterReconciler) SetupWithManager(ctx context.Context, mgr ctr
 			builder.WithPredicates(predicates.ClusterUnpaused(ctrl.LoggerFrom(ctx)))).
 		Complete(r)
 }
+
+func dhcpEnabled(cluster *infrav1alpha1.ProxmoxCluster) bool {
```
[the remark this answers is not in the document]
mcbenjemaa: yes

api/v1alpha1/proxmoxcluster_types.go
mcbenjemaa: but, this needs to be by default `false`

api/v1alpha1/helpers.go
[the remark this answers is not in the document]
mcbenjemaa: Please, do that in a separate PR.

api/v1alpha1/proxmoxcluster_types.go
avorima: You don't have to set a default. You can just treat nil like false

wikkyk:
Rebased and updated copyright headers.

api/v1alpha1/proxmoxcluster_types.go
```diff
@@ -50,25 +48,52 @@ type ProxmoxClusterSpec struct {
...
+	// +optional
+	Prefix int `json:"prefix,omitempty"`
+
+	// Gateway
+	// +optional
+	Gateway string `json:"gateway,omitempty"`
+
+	// DHCP indicates that if DHCP should be used to assign IP addresses.
```
wikkyk, line 90: DHCP indicates that ~if~ DHCP should be used to assign IP addresses.

internal/service/vmservice/bootstrap.go
```diff
@@ -40,9 +40,9 @@ func reconcileBootstrapData(ctx context.Context, machineScope *scope.MachineScop
 		return false, nil
 	}
 
-	if !machineHasIPAddress(machineScope.ProxmoxMachine) {
-		// skip machine doesn't have an IpAddress yet.
-		conditions.MarkFalse(machineScope.ProxmoxMachine, infrav1alpha1.VMProvisionedCondition, infrav1alpha1.WaitingForStaticIPAllocationReason, clusterv1.ConditionSeverityWarning, "no ip address")
+	if machineScope.VirtualMachine == nil {
+		// skip machine is not yet provisioned.
```
wikkyk, line 44: skip machine if it is not provisioned yet

wikkyk (changes requested):

wikkyk:
This PR should also update the docs as appropriate.

api/v1alpha1/proxmoxcluster_types.go
mcbenjemaa: so, i will have to change this fn and will be more complicated

```
// DHCPEnabled returns whether DHCP is enabled.
func (c ClusterNetworkConfig) DHCPEnabled() bool {
	switch {
	case (c.IPv6Config != nil && c.IPv6Config.DHCP) && (c.IPv4Config != nil && c.IPv4Config.DHCP):
		return true
	case (c.IPv6Config != nil && c.IPv6Config.DHCP) && c.IPv4Config == nil:
		return true
	case (c.IPv4Config != nil && c.IPv4Config.DHCP) && c.IPv6Config == nil:
		return true
	default:
		return false
	}
}

```

internal/service/vmservice/ip.go
```diff
@@ -74,13 +74,13 @@ func formatIPAddressName(name, device string) string {
 	return fmt.Sprintf("%s-%s", name, device)
 }
 
-func machineHasIPAddress(machine *infrav1alpha1.ProxmoxMachine) bool {
-	return machine.Status.IPAddresses[infrav1alpha1.DefaultNetworkDevice] != (infrav1alpha1.IPAddress{})
+func machineHasIPAddress(machine *infrav1.ProxmoxMachine) bool {
+	return machine.Status.IPAddresses[infrav1.DefaultNetworkDevice] != (infrav1.IPAddress{})
 }
 
 func handleIPAddressForDevice(ctx context.Context, machineScope *scope.MachineScope, device, format string, ipamRef *corev1.TypedLocalObjectReference) (string, error) {
```
wikkyk, line 81: Can this function be refactored so that it always returns an error when it also returns an empty string? To avoid all the `if err != nil || ip == "" {`.

internal/service/vmservice/vm.go
```diff
@@ -241,7 +242,9 @@ func reconcileMachineAddresses(scope *scope.MachineScope) error {
 	return nil
 }
 
-func getMachineAddresses(scope *scope.MachineScope) ([]clusterv1.MachineAddress, error) {
+// getMachineAddresses returns the ip addresses for the machine
+//nolint
```
wikkyk, line 246: why not?

wikkyk (changes requested):
First round of reviews.

api/v1alpha1/proxmoxcluster_types.go
avorima: Use `ptr.Deref` from `k8s.io/utils`

```go
// DHCPEnabled returns whether DHCP is enabled.
func (c ClusterNetworkConfig) DHCPEnabled() bool {
	switch {
	case (c.IPv6Config != nil && ptr.Deref(c.IPv6Config.DHCP, false)) && (c.IPv4Config != nil && ptr.Deref(c.IPv4Config.DHCP, false)):
		return true
	case (c.IPv6Config != nil && ptr.Deref(c.IPv6Config.DHCP, false)) && c.IPv4Config == nil:
		return true
	case (c.IPv4Config != nil && ptr.Deref(c.IPv4Config.DHCP, false)) && c.IPv6Config == nil:
		return true
	default:
		return false
	}
}
```

The bool pointer makes the DHCP setting clearer from the user's perspective because the explicit `false` stays `false`.

avorima: Actually, to reduce dependencies you can just copy the ptr.Deref function into the api package. Keeps imports clean

api/v1alpha1/proxmoxcluster_types.go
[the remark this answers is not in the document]
avorima: You're right, it's a bit out of scope for this PR. The webhook doesn't validate that IPv4 only gets prefix 32 though

internal/service/vmservice/ip.go
mcbenjemaa: It's very confusing to return an error when creating a new IP claim:

https://github.com/ionos-cloud/cluster-api-provider-proxmox/blob/0cd195bb54ec8ddd57c80061676d77e82fc1d6d0/internal/service/vmservice/ip.go#L94-L98

 also when trying to add a tag:

https://github.com/ionos-cloud/cluster-api-provider-proxmox/blob/0cd195bb54ec8ddd57c80061676d77e82fc1d6d0/internal/service/vmservice/ip.go#L112-L117

Another choice is to add a third return as bool, but it's actually the same.

api/v1alpha1/proxmoxcluster_types.go
mcbenjemaa: yes, you can check

.github/workflows/lint.yml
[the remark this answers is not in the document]
mcbenjemaa: https://github.com/ionos-cloud/cluster-api-provider-proxmox/pull/101

mcbenjemaa:
I'm ready

api/v1alpha1/proxmoxcluster_types.go
avorima: Ok. It can't be mutually exclusive if the addresses are always expected to have one item

internal/service/vmservice/vm.go
```diff
@@ -257,21 +259,70 @@ func getMachineAddresses(scope *scope.MachineScope) ([]clusterv1.MachineAddress,
...
+	var ipv4, ipv6 string
+	// get ip addresses from vm if dhcp is enabled
+	if hasDHCPEnabled(scope.InfraCluster.ProxmoxCluster.Spec.IPv4Config,
+		networkSpec.Default,
+		infrav1alpha1.DefaultNetworkDevice,
+		infrav1alpha1.IPV4Format) ||
+		hasDHCPEnabled(scope.InfraCluster.ProxmoxCluster.Spec.IPv6Config,
+			networkSpec.Default,
+			infrav1alpha1.DefaultNetworkDevice,
+			infrav1alpha1.IPV6Format) {
```
avorima, line 277: It would actually be less code if you move the default interface handling out of `hasDHCPEnabled` and use a few intermediate variables, e.g.

```go
if dev := networkSpec.Default; scope.InfraCluster.ProxmoxCluster.Spec.DHCPEnabled() || (dev != nil && dev.DHCP4 || dev.DHCP6) {
```

But more importantly, why is the default device DHCP setting even considered when DHCP is enabled on the cluster?

internal/service/vmservice/vm.go
```diff
@@ -257,21 +259,70 @@ func getMachineAddresses(scope *scope.MachineScope) ([]clusterv1.MachineAddress,
...
+		}
+
+		for _, net := range nets {
+			// default device
+			mac := extractMACAddress(scope.VirtualMachine.VirtualMachineConfig.Net0)
+			if strings.EqualFold(mac, net.HardwareAddress) {
+				for _, ip := range net.IPAddresses {
+					if ip.IPAddressType == "ipv4" && ip.IPAddress != scope.InfraCluster.ProxmoxCluster.Spec.ControlPlaneEndpoint.Host {
+						ipv4 = ip.IPAddress
+					} else if ip.IPAddressType == "ipv6" && !strings.Contains(ip.IPAddress, "fe80:") {
```
avorima, line 290: Please use [`netip.Addr.IsLinkLocalUnicast`](https://pkg.go.dev/net/netip#Addr.IsLinkLocalUnicast). Substring match for IPv6 is always less accurate than comparing the bits.

api/v1alpha1/proxmoxmachine_types.go
```diff
@@ -224,10 +224,20 @@ type NetworkDevice struct {
 	// +kubebuilder:validation:Minimum=1
 	// +kubebuilder:validation:Maximum=65520
 	MTU *uint16 `json:"mtu,omitempty"`
+
+	// DHCP4 indicates that if DHCP should be used to assign IPv4 addresses.
+	// DHCP4 enforce device to use DHCP despite the config set in cluster.spec.ipv4Config.
+	// +optional
+	DHCP4 bool `json:"dhcp4,omitempty"`
```
avorima, line 231: Why is in the base network device? All permuations of DHCP and IPAM on the default device should be configured in the cluster network config.

This also has the same pointer issue as the cluster dhcp setting.

api/v1alpha1/proxmoxcluster_types.go
```diff
@@ -50,25 +48,52 @@ type ProxmoxClusterSpec struct {
...
+	Prefix int `json:"prefix,omitempty"`
+
+	// Gateway
+	// +optional
+	Gateway string `json:"gateway,omitempty"`
+
+	// DHCP indicates that DHCP should be used to assign IP addresses.
+	// mutually exclusive with Addresses.
+	// +optional
+	// +kubebuilder:default=false
```
avorima, line 93: This default isn't needed. The behavior stays the same whether it's set to `false` or `null`. My pointer request was about making the user-facing API explicit and minimizing changes.

pkg/proxmox/goproxmox/api_client.go
```diff
@@ -245,3 +245,13 @@ func (c *APIClient) StartVM(ctx context.Context, vm *proxmox.VirtualMachine) (*p
 func (c *APIClient) TagVM(ctx context.Context, vm *proxmox.VirtualMachine, tag string) (*proxmox.Task, error) {
 	return vm.AddTag(ctx, tag)
 }
+
+// GetVMNetwork returns a VM network interfaces based on nodeName and vmID.
+func (c *APIClient) GetVMNetwork(ctx context.Context, vm *proxmox.VirtualMachine) (iFaces []*proxmox.AgentNetworkIface, err error) {
```
avorima, line 250: nit: named return values are not used and don't provide any additional context

avorima (commented):
Sorry, for the long wait.

api/v1alpha1/proxmoxcluster_types.go
mcbenjemaa: That's already considered via webhook.

api/v1alpha1/helpers.go
```diff
@@ -0,0 +1,168 @@
...
+	case (c.IPv6Config != nil && ptr.Deref(c.IPv6Config.DHCP, false)) && (c.IPv4Config != nil && ptr.Deref(c.IPv4Config.DHCP, false)):
+		return true
+	case (c.IPv6Config != nil && ptr.Deref(c.IPv6Config.DHCP, false)) && c.IPv4Config == nil:
+		return true
+	case (c.IPv4Config != nil && ptr.Deref(c.IPv4Config.DHCP, false)) && c.IPv6Config == nil:
+		return true
+	default:
+		return false
+	}
+}
```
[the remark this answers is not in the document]
mcbenjemaa: The problem here is we need to check if the cluster has dhcp so we can skip the IPAM pool.

Then, in the machine, we also need to check if the machine has DHCP.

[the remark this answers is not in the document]
avorima: This doesn't contradict my proposition

api/v1alpha1/proxmoxcluster_types.go
65278: Coming back to this, last week I debugged a setup which required static ip + dhcp. It's a valid usecase I guess.

avorima: I just noticed that my initial question came from a wrong assumption that the marker would force me to always provide addresses. The behavior is already how I would've wanted it to be. Sorry for the hold up

PR #54, opened 2024-01-08
Rename the environment variables to the same as in Packer

### LINKED ISSUE #52: PROXMOX_TOKEN in CAPMOX is PROXMOX_SECRET in the Image-Builder which is error-prone
I created an image with image-builder, the environment variable for the "secret-key" is `PROXMOX_TOKEN`. 
In CAPMOX the same environment variable contains the "token-user". 

During the clusterctl init the environment variable is preferred over the config-file, so i ended up to have an secret like this
```
apiVersion: v1
data:
  secret: M3OWRi
  token: M3OWRi
  url: aHqc29u
```
I would like to suggest to have the same naming as in the image-builder to save others the trouble shooting.
`PROXMOX_USERNAME` & `PROXMOX_TOKEN`

I should be able to create a PR for that - if desired.

3deep5me:
fixes #52 
**Draft** 
not tested (have to set up my development environment and tools first) 
*Description of changes:*
Renames the Variable `PROXMOX_TOKEN` to `PROXMOX_USERNAME`
Renames the Variable `PROXMOX_SECRET` to `PROXMOX_TOKEN`

This should be easier to understand and is the same as in Packer which should prevent #52.

*Testing performed:*

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -202,8 +202,8 @@ func initFlagsAndEnv(fs *pflag.FlagSet) {
 	klog.InitFlags(nil)
 
 	ProxmoxURL = env.GetString("PROXMOX_URL", "")
-	ProxmoxTokenID = env.GetString("PROXMOX_TOKEN", "")
-	ProxmoxSecret = env.GetString("PROXMOX_SECRET", "")
+	ProxmoxTokenID = env.GetString("PROXMOX_USERNAME", "")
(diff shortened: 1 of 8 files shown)


wikkyk:
I'm not so sure. Besides this being a breaking change, these are called Token ID and Secret in Proxmox VE. It would be confusing to call the secret the token.

PR #62, opened 2024-01-17
Add VRF device to support egress gateways

65278:
*Issue #, if available:*

*Description of changes:*
We require egress gateways for exposing services to the internet. VRF devices are a robust mechanism to do policy based routing. They also combine well with various BGP daemons and/or CNIs.

*Testing performed:*
Test cases added, clusters successfully rolled out.

*Discussion:*
The way networking is done in the machine template is rather wrong: We can not support multiple routes or IPs per network interface, or none of those if the devices is a slave of a virtual network devices. I've only added the pieces that are required to make this work, but I've designed it so that the existing code can be refactored easily.


api/v1alpha1/proxmoxmachine_types.go
```diff
@@ -203,6 +203,110 @@ type NetworkSpec struct {
...
+
+	// Routes are the routes associated with the l3mdev policy.
+	// +optional
+	// +kubebuilder:validation:MinItems=1
+	Routes []RouteSpec `json:"routes,omitempty"`
+
+	// RoutingPolicy is the l3mdev policy inserted into FiB.
+	// +optional
+	// +kubebuilder:validation:MinItems=1
+	RoutingPolicy []RoutingPolicySpec `json:"routing-policy,omitempty"`
```
mcbenjemaa, line 300: This is against K8s API Conventions,
Always use camel-case

mcbenjemaa: https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md

api/v1alpha1/proxmoxmachine_types.go
```diff
@@ -203,6 +203,110 @@ type NetworkSpec struct {
...
+}
+
+// VirtualNetworkDevices defines linux software networking devices.
+type VirtualNetworkDevices struct {
+	// Definition of a Vrf Device.
+	// +optional
+	// +listType=map
+	// +listMapKey=name
+	VRFs []VrfDevice `json:"vrfs,omitempty"`
```
mcbenjemaa, line 309: Why is this not part of additional devices?

mcbenjemaa: Can those VRFs be added with command-line?

api/v1alpha1/proxmoxmachine_types.go
```diff
@@ -203,6 +203,110 @@ type NetworkSpec struct {
 	// +listType=map
 	// +listMapKey=name
 	AdditionalDevices []AdditionalNetworkDevice `json:"additionalDevices,omitempty"`
+
+	// VirtualNetworkDevices defines virtual network devices (e.g. bridges, vlans ...).
+	VirtualNetworkDevices `json:",inline"`
+}
+
+// InterfaceConfig contains all configurables a network interface can have.
+type InterfaceConfig struct {
```
mcbenjemaa, line 212: Why do we need, IPV config here 😄

api/v1alpha1/proxmoxmachine_types.go
65278: Good point. My mind was copying netplan there.

api/v1alpha1/proxmoxmachine_types.go
65278: That's because the rules for validation would need reworking. Additional devices are strictly "physical" at the moment.
It makes sense to split the setup into physical and software defined, as this is effectively how netplan handles it too.

Can VRFs be added by CLI? Yes. Unfortunately, we do not have the information to do so. This requires knowing the interface name. We do not know the actual interface name until after the machine boots, as this is not deterministic (interface names are made from rules matching pci-e addresses). Netplan arranges those things for us via mac address matching.

The short gist of it is: we're using netplan and we're not reinventing the wheel.

api/v1alpha1/proxmoxmachine_types.go
65278: Because virtual devices can have IPs, and I'm not re-adding that block for each device type that I add (bridges, l2bonds, vlans ...). In fact I'm going to refactor the interface logic a little bit so all devices can have routes added, and this is a stub for doing so. The best thing we can do is to most directly map netplan and reduce anything

docs/Usage.md
```diff
@@ -95,14 +95,24 @@ IP_PREFIX: "25"                                               # Subnet Mask in C
...
-
-EXP_CLUSTER_RESOURCE_SET: "true"                               # This enables the ClusterResourceSet feature that we are using to deploy CNI
+## -- xl nodes -- ##
+BOOT_VOLUME_DEVICE: "scsi0"                                   # The device used for the boot disk.
+BOOT_VOLUME_SIZE: "100"                                       # The size of the boot disk in GB.
+NUM_SOCKETS: "2"                                              # The number of sockets for the VMs.
+NUM_CORES: "4"                                                # The number of cores for the VMs.
+MEMORY_MIB: "8048"                                            # The memory size for the VMs.
+
+## -- loadbalancer nodes -- #
```
mcbenjemaa, line 105: I would say the docs about load balancing can go to the advanced-setups.md instead.

Makefile
```diff
@@ -227,6 +227,24 @@ CALICO_VERSION ?= v3.26.3
 crs-calico: ## Generates crs manifests for Calico.
 	curl -o templates/crs/cni/calico.yaml https://raw.githubusercontent.com/projectcalico/calico/$(CALICO_VERSION)/manifests/calico.yaml
 
+METALLB_VERSION ?= main
+FRR_K8S_DIR = metallb/charts/metallb/charts/frr-k8s/templates
+.PHONY: crs-metallb
+crs-metallb: ## Generates crs manifests for MetalLB.
+	#@ TODO: temporarily pull metallb from main, update when release happens
+	@rm -rf metallb
+	@git clone https://github.com/metallb/metallb/ metallb/ -n -q -b main --single-branch --filter=blob:none
```
mcbenjemaa, line 236: Is there a reason you're cloning the repo?

You can use the helm repo
```
helm repo add metallb https://metallb.github.io/metallb
helm install metallb metallb/metallb
```

api/v1alpha1/proxmoxmachine_types.go
```diff
@@ -203,6 +203,110 @@ type NetworkSpec struct {
...
+	// +optional
+	To string `json:"to,omitempty"`
+
+	// From is the subnet of the source.
+	// +optional
+	From string `json:"from,omitempty"`
+
+	// Table is the routing table id.
+	// +optional
+	Table uint32 `json:"table,omitempty"`
```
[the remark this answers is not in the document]
65278: It's called table because that's what it is. We could call it TableId, but that's not too helpful either. If you do not know what it is, then you won't need it. If you need it, you know what it is.
http://linux-ip.net/html/routing-tables.html

pkg/cloudinit/network.go
```diff
@@ -57,7 +59,48 @@ const (
...
         {{- end -}}
       {{- end -}}
-  {{- end -}}`
+  {{- end -}}
+  {{- end -}}
+  {{- $vrf := 0 -}}
+  {{- range $index, $element := .NetworkConfigData }}
+  {{- if eq $element.Type "vrf" }}
+  {{- if eq $vrf 0 }}
+  vrfs:
```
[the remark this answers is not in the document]
65278: I was considering this, but all things will eventually be template blocks, and then we'll have to copy those template blocks to each template.

docs/Usage.md
65278: Alright.

api/v1alpha1/proxmoxmachine_types.go
```diff
@@ -203,6 +203,110 @@ type NetworkSpec struct {
...
+
+// VrfDevice defines Virtual Routing Flow devices.
+type VrfDevice struct {
+	// Interfaces is the list of proxmox network devices managed by this virtual device.
+	Interfaces []string `json:"interfaces,omitempty"`
+
+	// Name is the virtual network device name.
+	// must be unique within the virtual machine.
+	// +kubebuilder:validation:MinLength=3
+	Name string `json:"name"`
```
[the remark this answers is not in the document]
65278: This is a freeform field and has no relation to existing interfaces, as it's the name of the to-be-created interface. There's a point to be made of this being useless, but I think it would come in handy when we need to add interface specific configuration to frr-k8s. What we could do is generate a name unless you give it one.

Makefile
65278: Unfortunately the route import feature is broken in metallb 0.13.12, so I've been talking with the MetalLB guys and they guaranteed to me that it's fixed in the next release and that release will be out soon.

In fact the release just dropped: https://metallb.universe.tf/release-notes/
I'll be removing most of this.

65278: Done

api/v1alpha1/proxmoxmachine_types.go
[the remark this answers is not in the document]
65278: I've done this, but now VRFDevices are no longer a map, meaning that you can happily create two devices with the same name. Not certain that's an improvement.

api/v1alpha1/proxmoxmachine_types.go
```diff
@@ -203,6 +203,110 @@ type NetworkSpec struct {
...
+
+	// Priority is the position in the ip rule fib table.
+	// +kubebuilder:validation:Maximum=4294967295
+	// +kubebuilder:validation:XValidation:message="Cowardly refusing to insert fib rule matching kernel rules",rule="(self > 0 && self < 32765) || (self > 32766)"
+	// +optional
+	Priority uint32 `json:"priority,omitempty"`
+}
+
+// VrfDevice defines Virtual Routing Flow devices.
+type VrfDevice struct {
```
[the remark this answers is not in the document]
65278: Done

api/v1alpha1/proxmoxmachine_types.go
65278: Netplan has a section for things you can set on every kind of device (or even outwith of devices):
https://netplan.readthedocs.io/en/stable/netplan-yaml/#properties-for-all-device-types
Everything that goes into that section we'd duplicate per network device kind. Therefore we make one datastructure that will hold all possible configurables. For example: `mtu, dhcp, dhcpv6, nameservers` ...

Doing that requires a bigger refactor on the codebase, so I haven't done this here, but that's what this datastructure is for.

Makefile
```diff
@@ -227,6 +227,17 @@ CALICO_VERSION ?= v3.26.3
 crs-calico: ## Generates crs manifests for Calico.
 	curl -o templates/crs/cni/calico.yaml https://raw.githubusercontent.com/projectcalico/calico/$(CALICO_VERSION)/manifests/calico.yaml
 
+METALLB_VERSION ?= 0.14.3
+FRR_K8S_DIR = metallb/charts/metallb/charts/frr-k8s/templates
+METALLB_TOLERATIONS = [{"key": "node-role.kubernetes.io/load-balancer", "operator": "Exists", "effect": "NoSchedule"}]
+.PHONY: crs-metallb
+crs-metallb: ## Generates crs manifests for MetalLB.
+	$(HELM) repo add metallb https://metallb.github.io/metallb
```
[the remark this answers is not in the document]
65278: Done. Should also do that for crs-cilium

api/v1alpha1/proxmoxmachine_types.go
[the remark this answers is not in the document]
65278: After monkeying around with validation, there's no good way to guarantee uniqueness when we remove the map. I'm reverting this, because having to enter interface names here is the less bad option.

Problems with validation:
- Checking that the entered interface name does not collide with an autogenerated interface name is too expensive:
  `Forbidden: estimated rule cost exceeds budget by factor of more than 100x (try simplifying the rule, or adding maxItems, maxProperties, and maxLength where arrays, maps, and strings are declared`
- Checking that the VRF list items are unique is also too expensive:
  `Forbidden: uniqueItems cannot be set to true since the runtime complexity becomes quadratic`

docs/Usage.md
65278: Done

pkg/cloudinit/network.go
```diff
@@ -89,9 +132,23 @@ func (r *NetworkConfig) validate() error {
 		return ErrMissingNetworkConfigData
 	}
 	for _, d := range r.data.NetworkConfigData {
+		// refactor this when network configuration is unified
```
wikkyk, line 135: please prepend `TODO` to help sonarcloud keep track of it

wikkyk (changes requested):
Just this one thing, then as far as I'm concerned it's good to go. Don't forget to get a successful e2e run before merging.

pkg/cloudinit/network.go
65278: Done. I'll also make an issue

wikkyk (approved):
Please don't forget to rebase and run e2e.

wikkyk (approved):

65278:
The e2e tests failing has nothing to do with the PR itself. It fails on capmox-controller-manager not starting. The same e2e test works with this PR using my notebook.

PR #74, opened 2024-01-22
Disable licence header check

wikkyk:
*Issue #, if available:*
https://github.com/ionos-cloud/cluster-api-provider-proxmox/actions/runs/7612075368/job/20728874326

*Description of changes:*
The check does not work very well with multiple years.
I might build a better one sometime. A good check should only look at the files changed in the PR and make sure the year/s is/are correctly updated.

Note: update the branch protection rule prior to merging.


mcbenjemaa:
what about old files

mcbenjemaa (approved):

PR #75, opened 2024-01-23
fix: remove boilerplate.go.txt

wikkyk:
The copyright status of autogenerated files is at best dubious. The copyright year does not change just because the calendar year has changed. The header will only cause problems every January.


Makefile
```diff
@@ -46,7 +46,7 @@ manifests: controller-gen ## Generate WebhookConfiguration, ClusterRole and Cust
 
 .PHONY: generate
 generate: controller-gen ## Generate code containing DeepCopy, DeepCopyInto, and DeepCopyObject method implementations.
-	$(CONTROLLER_GEN) object:headerFile="hack/boilerplate.go.txt" paths="./..."
+	$(CONTROLLER_GEN) object paths="./..."
```
mcbenjemaa, line 49: Why not create another special boilerplate go autogenerated.

wikkyk:
depends on #76

Makefile
wikkyk: What for?

mcbenjemaa: copyright 😄

wikkyk: Autogenerated files aren't copyrightable.

mcbenjemaa: can you put the boilerplate file empty, and keep this command as it is?

wikkyk: I could but why would we keep an empty file around?

mcbenjemaa (approved):

Makefile
avorima: > Autogenerated files aren't copyrightable.

I must disagree here. All files that are compiled to should have the license header even those that are auto-generated. It is not required that the year be kept up to date, so rather the validation that the year matches should be loosened

PR #76, opened 2024-01-23
fix(ci): fold verify under go_test

wikkyk:
`verify` running in `pull_request_target` context verifies the upstream rather than the PR.

*Issue #, if available:*
https://github.com/ionos-cloud/cluster-api-provider-proxmox/actions/runs/7627247798/job/20775470951

*Description of changes:*
Moved `verify` to `go_test`, as `verify` also needs to checkout the PR. Whilst it's possible for those two to run in parallel, in practice `verify` is very fast anyway and we're avoiding copying the checkout dance which wouldn't be good for maintainability.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -7,15 +7,6 @@ on:
     branches: ["main"]
 
 jobs:
-  verify:
-    runs-on: ubuntu-latest
-    steps:
-      - uses: actions/checkout@v4.1.1
-      - uses: actions/setup-go@v4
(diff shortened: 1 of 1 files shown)


wikkyk:
Note `verify` will fail here as it runs the old logic (which is the issue being fixed).

wikkyk:
Because it's running from main instead of the PR branch, which is the exact issue being fixed by this PR.

mcbenjemaa (approved):

PR #79, opened 2024-01-23
fix(cloudinit): wrap ipv6 addresses in quotation marks

lucasl0st:
The ipv6 gateway needs to be wrapped in quotation marks, otherwise it can produce invalid yaml.

Example (invalid, caused by ::, ::1 is fine, can be checked by a yaml linter like https://www.yamllint.com):

```yaml
network:
  version: 2
  renderer: networkd
  ethernets:
    eth0:
      match:
        macaddress: BC:24:11:75:4E:D2
      dhcp4: false
      dhcp6: false
      addresses:
        - 2a01:7e01:e003:3b04:ffff::/64
      routes:
        - to: '::/0'
          via: 2a01:7e01:e003:3b04::
      nameservers:
        addresses:
          - 2a01:7e01:e003:3b00::6464
 ```

Results into:

image

### DIFF
--- a/pkg/cloudinit/network.go
+++ b/pkg/cloudinit/network.go
@@ -38,7 +38,7 @@ const (
         - {{ $element.IPAddress }}
       {{- end }}
       {{- if and $element.IPV6Address (not $element.DHCP6)}}
-        - {{ $element.IPV6Address }}
+        - '{{ $element.IPV6Address }}'
 	  {{- end }}
       routes:
       {{- if and $element.Gateway (not $element.DHCP4) }}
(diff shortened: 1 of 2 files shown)


avorima (commented):
Doesn't this apply to all ipv6 addresses in the template?

lucasl0st:
> Doesn't this apply to all ipv6 addresses in the template?

I first thought no but yes you are right. An address with just `::` at the end causes some interesting behaviour when parsed:

image

avorima:
Please re-title the PR

avorima (approved):

lucasl0st:
Changed the tile, external contributors don't have permissions to merge btw, just in case you are waiting for me

mcbenjemaa:
Thanks for addressing this

avorima:
TIL

PR #85, opened 2024-01-24
ci: fix sonarcloud scan pr source

lpape-ionos:
Sonarcloud thinks that every branch is "main" (see https://sonarcloud.io/project/branches_list?id=ionos-cloud_cluster-api-provider-proxmox)

This is an attempt to fix that

Warning: the check run here uses the workflow from main, so I cant be sure if this works. I'd be happy to receive other suggestions

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -38,8 +38,21 @@ jobs:
       - name: Run tests
         run: "make test"
 
-      - name: SonarCloud Scan
+      - name: SonarCloud Scan ${{ github.ref }}
         uses: SonarSource/sonarcloud-github-action@v2.1.1
+        if: ${{ github.event_name == 'push' }}
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}  # Needed to get PR information, if any
+          SONAR_TOKEN: ${{ secrets.SONAR_TOKEN }}
+
+      # If triggered by a PR, we have to use the PR's source
+      - name: SonarCloud Scan (preview) merge commit for PR ${{ github.event.pull_request.number }}
+        uses: SonarSource/sonarcloud-github-action@v2.1.1
+        if: ${{ github.event_name == 'pull_request_target' }}
+        with:
+         args: >
+           -Dsonar.pullrequest.key=${{ github.event.pull_request.number }}
+           -Dsonar.scm.revision=${{ github.event.pull_request.head.sha }}
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}  # Needed to get PR information, if any
           SONAR_TOKEN: ${{ secrets.SONAR_TOKEN }}


avorima:
There are a few things to consider here.

I guess you found [this page](https://docs.sonarsource.com/sonarcloud/advanced-setup/ci-based-analysis/other-cis/), but [another page](https://docs.sonarsource.com/sonarcloud/enriching/branch-analysis-setup/) mentions `sonar.branch.name` and `sonar.branch.target`. Then the [github action setup docs](https://docs.sonarsource.com/sonarcloud/advanced-setup/ci-based-analysis/github-actions-for-sonarcloud/) don't mention any of these.

The scan action seems to have access to `GITHUB_EVENT_NAME`, but also `GITHUB_BASE_REF`, `GITHUB_HEAD_REF` and `GITHUB_REF`. That should be enough to automatically fill out the required information.

Here are action logs showing the event name:
```
INFO: Load project pull requests
INFO: Load project pull requests (done) | time=205ms
INFO: Load branch configuration
INFO: Github event: pull_request_target
INFO: Load branch configuration (done) | time=3ms
```

Then there's this issue https://github.com/SonarSource/sonarqube-scan-action/issues/29 which also mentions that `sonar.pullrequest.*` **should** be populated.

I tried looking finding where these variables are used, but couldn't find anything at first glance in https://github.com/SonarSource/sonar-scanner-cli or https://github.com/SonarSource/sonar-scanner-commons

I'm not saying your change won't work, but I think the culprit actually is sonarcloud here. So I'd propose that you open an issue (or reopen the one I linked).

lpape-ionos:
I agree with you that sonar should already have all the necessary information but I dont think that this is sonar's fault. We are doing an unconventional checkout here: https://github.com/ionos-cloud/cluster-api-provider-proxmox/blob/01b3f7f6dc8c27aa0f50727401c1c14c818aa402/.github/workflows/test.yml#L22-L29

This is a workaround for branches across forks, my theory is that this confuses the sonar scan. I don't know how one would solve that properly

wikkyk:
Old thread but it doesn't seem like much has changed since and sonarcloud still doesn't support `pull_request_target`:
https://community.sonarsource.com/t/github-comments-stopped-working-in-github-action-after-switch-to-pull-request-target/42556/12

.github/workflows/test.yml
```diff
@@ -38,22 +38,12 @@ jobs:
...
-         args: >
-           -Dsonar.pullrequest.key=${{ github.event.pull_request.number }}
-           -Dsonar.pullrequest.branch=${{ github.event.pull_request.head.ref }}
-           -Dsonar.pullrequest.base=${{ github.event.pull_request.base.ref }}
-        env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}  # Needed to get PR information, if any
-          SONAR_TOKEN: ${{ secrets.SONAR_TOKEN }}
+          args: >
+            -Dsonar.pullrequest.key=${{ github.event.pull_request.number }}
+            -Dsonar.scm.revision=${{ github.event.pull_request.head.sha }}
```
wikkyk, line 49: I think this should be sufficient.

avorima:
> Old thread but it doesn't seem like much has changed since and sonarcloud still doesn't support `pull_request_target`: https://community.sonarsource.com/t/github-comments-stopped-working-in-github-action-after-switch-to-pull-request-target/42556/12

Yeah, that's what I figured. I thought that opening an issue would shed some light on this

avorima (approved):

lpape-ionos:
Looks good to me, tested in a test-PR: https://github.com/ionos-cloud/cluster-api-provider-proxmox/pull/91

PR #93, opened 2024-01-26
sonar: fix paths

wikkyk:
The sonar config is broken in several ways:

1. it's sonar.**test**.{in,ex}clusions. not sonar.**tests**.{in,ex}clusions
2. unset sonar.tests disables the test scope altogether
3. fixing the above still leaves us with incorrect {in,ex}clusions lists.

https://docs.sonarsource.com/sonarqube/latest/project-administration/analysis-scope/ has a neat example that I more or less extended.


avorima (commented):
The page you linked is quite interesting. Can you add `sonar.coverage.exclusions=**/main.go`?

wikkyk:
I could but I don't see why shouldn't `main.go`s not count towards coverage%. [cmd/main.go](https://github.com/ionos-cloud/cluster-api-provider-proxmox/blob/main/cmd/main.go) is authored by humans, it has logic, that logic is non-trivial, and it's currently untested.

mcbenjemaa (approved):
we can also have unit test for main.go

PR #101, opened 2024-02-01
Fix Lint action

mcbenjemaa:
The lint action is broken, due to the update

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -1,6 +1,10 @@
 name: Lint
 on: [ pull_request ]
 
+permissions:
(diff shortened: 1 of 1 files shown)


65278 (approved):
LGTM.
The only thing I'd like is a more exhaustive description of the issue fixed, in the future.

From what I gather, golangci-lint wasn't broken, but the setup phase printed about 40000 lines of warnings while unpacking a tarball. That's certainly worth addressing.

mcbenjemaa:
it's a version conflicts issue

wikkyk:
merging, no need for e2e test

Issue #95, opened 2024-01-28
Support externally managed Kubernetes Control Planes


prometherion:
**Describe the solution you'd like**

Cluster API supports referencing a third-party Control Plane Provider, such as [Kamaji](https://github.com/clastix/cluster-api-control-plane-provider-kamaji).

The Control Plane provider will be responsible for provisioning a Control Plane backed by a Kubernetes API Server which will be used by the infrastructure cluster with the `controlPlaneEndpoint` contract.

The current version of CAPMOX doesn't consider this, requiring to know the Control Plane endpoint in advance.

https://github.com/ionos-cloud/cluster-api-provider-proxmox/blob/0033e2954e5fc70029fe64fae1daf641a24dcd2b/internal/webhook/proxmoxcluster_webhook.go#L99-L108

Kamaji has successfully integrated with other infrastructure providers by relying on externally managed Control Planes: this approach has been taken into consideration on other CAPI infrastructure providers we worked with.

**Anything else you would like to add:**

To avoid breaking the current UX, it could be useful to have a knob to skip the ControlPlane endpoint address for the given `ProxmoxCluster`.

Once the Control Plane provider has provisioned the Kubernetes API Server address, it can patch the infrastructure cluster to continue the required reconciliation, e.g.: https://github.com/clastix/cluster-api-control-plane-provider-kamaji/blob/75b0578114b236f1741d4b79b60eb39b23dfcbeb/controllers/kamajicontrolplane_controller_cluster_patch.go#L20-L52

I'm open to provide a PR to address this feature request.

**Environment:**

- Cluster-api-provider-proxmox version: N.R.
- Kubernetes version: (use `kubectl version`): N.R.
- OS (e.g. from `/etc/os-release`):  N.R.

mcbenjemaa:
Thanks for the issue, 

you're right; we will take this into our next release.

prometherion:
May I work on this? I'd like to contribute to the project.

wikkyk:
fixed by #106

PR #106, opened 2024-02-09
feat: making control plane endpoint fields optional

prometherion:
*Issue #95*

*Description of changes:*

Skipping webhook validation for the `ProxmoxCluster` Control Plane endpoint field.

The `ProxmoxCluster` controller will wait for a valid IP before proceeding to mark the infrastructure ready.

*Testing performed:*

N.A.

### DIFF
--- a/internal/controller/proxmoxcluster_controller.go
+++ b/internal/controller/proxmoxcluster_controller.go
@@ -168,6 +168,31 @@ func (r *ProxmoxClusterReconciler) reconcileNormal(ctx context.Context, clusterS
 	// If the ProxmoxCluster doesn't have our finalizer, add it.
 	ctrlutil.AddFinalizer(clusterScope.ProxmoxCluster, infrav1alpha1.ClusterFinalizer)
 
+	cpe := clusterScope.ControlPlaneEndpoint()
+	if cpe.Port == 0 {
+		conditions.MarkFalse(
+			clusterScope.ProxmoxCluster,
+			infrav1alpha1.ProxmoxClusterReady,
+			infrav1alpha1.MissingControlPlaneEndpointPortReason,
+			clusterv1.ConditionSeverityWarning,
+			"The ProxmoxCluster is missing or waiting for a ControlPlaneEndpoint port",
+		)
+
+		return reconcile.Result{}, nil
+	}
+
+	if cpe.Host == "" {
+		conditions.MarkFalse(
+			clusterScope.ProxmoxCluster,
+			infrav1alpha1.ProxmoxClusterReady,
+			infrav1alpha1.MissingControlPlaneEndpointHostReason,
+			clusterv1.ConditionSeverityWarning,
+			"The ProxmoxCluster is missing or waiting for a ControlPlaneEndpoint hostname",
+		)
+
+		return reconcile.Result{}, nil
+	}
+
 	res, err := r.reconcileIPAM(ctx, clusterScope)
 	if err != nil {
 		return ctrl.Result{}, err
--- a/internal/webhook/proxmoxcluster_webhook.go
+++ b/internal/webhook/proxmoxcluster_webhook.go
@@ -94,64 +94,64 @@ func (*ProxmoxCluster) ValidateUpdate(_ context.Context, _ runtime.Object, newOb
 func validateIPs(cluster *infrav1.ProxmoxCluster) error {
 	ep := cluster.Spec.ControlPlaneEndpoint
 
-	gk, name := cluster.GroupVersionKind().GroupKind(), cluster.GetName()
+	if ep.Host != "" && ep.Port > 0 {
+		gk, name := cluster.GroupVersionKind().GroupKind(), cluster.GetName()
 
-	ipAddr, err := netip.ParseAddrPort(fmt.Sprintf("%s:%d", ep.Host, ep.Port))
-	if err != nil {
-		return apierrors.NewInvalid(
-			gk,
-			name,
-			field.ErrorList{
-				field.Invalid(
-					field.NewPath("spec", "controlplaneEndpoint"), fmt.Sprintf("%s:%d", ep.Host, ep.Port), "provided endpoint is not in a valid IP and port format"),
-			})
-	}
-
-	// IPv4
-	if cluster.Spec.IPv4Config != nil {
-		set, err := buildSetFromAddresses(cluster.Spec.IPv4Config.Addresses)
+		ipAddr, err := netip.ParseAddrPort(fmt.Sprintf("%s:%d", ep.Host, ep.Port))
 		if err != nil {
 			return apierrors.NewInvalid(
 				gk,
 				name,
 				field.ErrorList{
 					field.Invalid(
-						field.NewPath("spec", "IPv4Config", "addresses"), cluster.Spec.IPv4Config.Addresses, "provided addresses are not valid IP addresses, ranges or CIDRs"),
+						field.NewPath("spec", "controlplaneEndpoint"), fmt.Sprintf("%s:%d", ep.Host, ep.Port), "provided endpoint is not in a valid IP and port format"),
 				})
 		}
+		// IPv4
+		if cluster.Spec.IPv4Config != nil {
+			set, err := buildSetFromAddresses(cluster.Spec.IPv4Config.Addresses)
+			if err != nil {
+				return apierrors.NewInvalid(
+					gk,
+					name,
+					field.ErrorList{
+						field.Invalid(
+							field.NewPath("spec", "IPv4Config", "addresses"), cluster.Spec.IPv4Config.Addresses, "provided addresses are not valid IP addresses, ranges or CIDRs"),
+					})
+			}
 
-		if set.Contains(ipAddr.Addr()) {
-			return apierrors.NewInvalid(
-				gk,
-				name,
-				field.ErrorList{
-					field.Invalid(
-						field.NewPath("spec", "IPv4Config", "addresses"), cluster.Spec.IPv4Config.Addresses, "addresses may not contain the endpoint IP"),
-				})
+			if set.Contains(ipAddr.Addr()) {
+				return apierrors.NewInvalid(
+					gk,
+					name,
+					field.ErrorList{
+						field.Invalid(
+							field.NewPath("spec", "IPv4Config", "addresses"), cluster.Spec.IPv4Config.Addresses, "addresses may not contain the endpoint IP"),
+					})
+			}
 		}
-	}
+		// IPV6
+		if cluster.Spec.IPv6Config != nil {
+			set6, err := buildSetFromAddresses(cluster.Spec.IPv6Config.Addresses)
+			if err != nil {
+				return apierrors.NewInvalid(
+					gk,
+					name,
+					field.ErrorList{
+						field.Invalid(
+							field.NewPath("spec", "IPv6Config", "addresses"), cluster.Spec.IPv6Config.Addresses, "provided addresses are not valid IP addresses, ranges or CIDRs"),
+					})
+			}
 
-	// IPV6
-	if cluster.Spec.IPv6Config != nil {
-		set6, err := buildSetFromAddresses(cluster.Spec.IPv6Config.Addresses)
-		if err != nil {
-			return apierrors.NewInvalid(
-				gk,
-				name,
-				field.ErrorList{
-					field.Invalid(
-						field.NewPath("spec", "IPv6Config", "addresses"), cluster.Spec.IPv6Config.Addresses, "provided addresses are not valid IP addresses, ranges or CIDRs"),
-				})
-		}
-
-		if set6.Contains(ipAddr.Addr()) {
-			return apierrors.NewInvalid(
-				gk,
-				name,
-				field.ErrorList{
-					field.Invalid(
-						field.NewPath("spec", "IPv6Config", "addresses"), cluster.Spec.IPv6Config.Addresses, "addresses may not contain the endpoint IP"),
-				})
+			if set6.Contains(ipAddr.Addr()) {
+				return apierrors.NewInvalid(
+					gk,
+					name,
+					field.ErrorList{
+						field.Invalid(
+							field.NewPath("spec", "IPv6Config", "addresses"), cluster.Spec.IPv6Config.Addresses, "addresses may not contain the endpoint IP"),
+					})
+			}
 		}
 	}
 
--- a/api/v1alpha1/conditions_consts.go
+++ b/api/v1alpha1/conditions_consts.go
@@ -83,4 +83,7 @@ const (
 const (
 	// ProxmoxClusterReady documents the status of ProxmoxCluster and its underlying resources.
 	ProxmoxClusterReady clusterv1.ConditionType = "ClusterReady"
+	// Status of the ProxmoxCluster control plane endpoint fields.
+	MissingControlPlaneEndpointPortReason = "MissingControlPlaneEndpointPort"
+	MissingControlPlaneEndpointHostReason = "MissingControlPlaneEndpointHost"
 )


prometherion:
Before providing more test coverage, may I ask for a simple review of the proposed changes, from a workflow perspective?

It's not clear to me if the maintainers are open to skipping the validation of the `ProxmoxCluster.spec.controlPlaneEnpdoint` field.

avorima:
Hi @prometherion, thank you for your contribution. Skipping validation for optional fields is fine, but please make sure your changes don't alter the validation behavior of other fields. Users should see errors as early as possible.
The controller checks for empty control plane endpoint should be done after IPAM was reconciled, so that they're not waiting for the external process to set the control plane endpoint.

Does Kamaji set the port and address separately? I'm asking, because my understanding was that the endpoint is always written in full or not at all. This would mean that you could merge the two conditions into one `MissingControlPlaneEndpoint`.

prometherion:
> Does Kamaji set the port and address separately?

Kamaji is doing in a single transaction, yes, I can uniform those checks.

mcbenjemaa:
take your time mate

mcbenjemaa:
@prometherion 
Are there any updates on this, or can I try it

prometherion:
Finally, I'm revamping it, feeling sorry for being late @mcbenjemaa.

Let me know if we're ready to get this tested.

mcbenjemaa:
can you provide a use-case to test it with Kamaji?

internal/controller/proxmoxcluster_controller.go
```diff
@@ -168,6 +168,14 @@ func (r *ProxmoxClusterReconciler) reconcileNormal(ctx context.Context, clusterS
 	// If the ProxmoxCluster doesn't have our finalizer, add it.
 	ctrlutil.AddFinalizer(clusterScope.ProxmoxCluster, infrav1alpha1.ClusterFinalizer)
 
+	cpe := clusterScope.ControlPlaneEndpoint()
+	switch {
+	case cpe.Host == "":
+		conditions.MarkFalse(clusterScope.ProxmoxCluster, infrav1alpha1.ProxmoxClusterReady, infrav1alpha1.MissingControlPlaneEndpointReason, clusterv1.ConditionSeverityWarning, "The ProxmoxCluster is missing or waiting for a ControlPlaneEndpoint host")
```
avorima, line 174: This information is essentially only visible until IPAM is reconciled, which if IIRC doesn't take more than a few seconds. It would probably be good to also log and/or create an event for either of these.

prometherion: Thanks, just rebased and added the logs.

I skipped events generation since I don't know how they should be formatted since there aren't events generated in the provider.

wikkyk (changes requested):
I think this is close to done. The core is there but I'm not sold on the details.. This could use some test cases which would've uncovered the inconsistency I pointed out inline.

However, I don't think I like the 'magic' behaviour of empty `host` and(?) `port` meaning externally managed. Someone could omit host/port by accident without actually intending to use Kamaji or so. This needs very clear documentation. (Technically, this doesn't actually make the fields optional (-:)

Personally, I would prefer an optional bool field in `ProxmoxClusterSpec` like `ControlPlaneExternallyManaged` and to require that either `ControlPlaneExternallyManaged` or `ControlPlaneEndpoint` is set. This would make the intent clear.

prometherion:
> However, I don't think I like the 'magic' behaviour of empty `host` and(?) `port` meaning externally managed. Someone could omit host/port by accident without actually intending to use Kamaji or so. This needs very clear documentation. (Technically, this doesn't actually make the fields optional (-:)

Documentation is a great place, of course, wondering if we could implement this kind of check in a different way.

The Cluster API has a contract for an externally managed control plane thanks to the status key [`externalManagedControlPlane`](https://cluster-api.sigs.k8s.io/developer/architecture/controllers/control-plane#optional-status-fields).

When a user is open to using Kamaji, or any other Control Plane provider which is satisfying that status key contract, we could skip the validation requiring a filled `ControlPlaneEndpoint` struct. With this, we're creating more guardrails for users that otherwise could shoot themselves in the foot.

internal/webhook/proxmoxcluster_webhook.go
```diff
@@ -24,14 +24,15 @@ import (
...
-	infrav1 "github.com/ionos-cloud/cluster-api-provider-proxmox/api/v1alpha1"
 	"github.com/pkg/errors"
 	"go4.org/netipx"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/util/validation/field"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
+
+	infrav1 "github.com/ionos-cloud/cluster-api-provider-proxmox/api/v1alpha1"
```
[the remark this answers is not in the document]
prometherion: It's my IDE sorting go imports, I'll amend it.

wikkyk:
Two points, then:
1. We should set [externalManagedControlPlane](https://cluster-api.sigs.k8s.io/developer/architecture/controllers/control-plane#optional-status-fields) in the status when selected
2. I like it when things are explicit (or I don't like magic values) so I'd opt for a new key in `ProxmoxClusterSpec` (we can name it `ExternalManagedControlPlane` to stay consistent) and the user either sets it to `true` or populates `ControlPlaneEndpoint` and we validate that only one of those is set.

prometherion:
> 1. We should set [externalManagedControlPlane](https://cluster-api.sigs.k8s.io/developer/architecture/controllers/control-plane#optional-status-fields) in the status when selected

A bit confused, here. For our controller (Kamaji) it's already done, nothing's expected from the Proxmox CAPI provider.

> 2\. I like it when things are explicit (or I don't like magic values) so I'd opt for a new key in `ProxmoxClusterSpec` (we can name it `ExternalManagedControlPlane` to stay consistent) and the user either sets it to `true` or populates `ControlPlaneEndpoint` and we validate that only one of those is set.

Working on this.

internal/webhook/proxmoxcluster_webhook.go
```diff
@@ -94,6 +95,11 @@ func (*ProxmoxCluster) ValidateUpdate(_ context.Context, _ runtime.Object, newOb
 
 func validateControlPlaneEndpoint(cluster *infrav1.ProxmoxCluster) error {
 	ep := cluster.Spec.ControlPlaneEndpoint
+	// Skipping the validation of the Control Plane endpoint in case of an empty value:
+	// this is the case of externally managed Control Plane which eventually provides the LB.
+	if ep.Host == "" && ep.Port == 0 {
```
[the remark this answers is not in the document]
prometherion: Outdated, with the new flag no need anymore for this check.

internal/webhook/proxmoxcluster_webhook.go
```diff
@@ -94,6 +95,11 @@ func (*ProxmoxCluster) ValidateUpdate(_ context.Context, _ runtime.Object, newOb
...
+		return nil
+	}
```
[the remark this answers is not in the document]
prometherion: Although outdated, we're keeping up with Cluster conditions.

prometherion:
Dunno who to ping here, I think I addressed all the required changes.

internal/controller/proxmoxcluster_controller.go
```diff
@@ -168,6 +168,23 @@ func (r *ProxmoxClusterReconciler) reconcileNormal(ctx context.Context, clusterS
 	// If the ProxmoxCluster doesn't have our finalizer, add it.
 	ctrlutil.AddFinalizer(clusterScope.ProxmoxCluster, infrav1alpha1.ClusterFinalizer)
 
+	if clusterScope.ProxmoxCluster.Spec.ExternalManagedControlPlane {
+		switch {
```
wikkyk, line 172: I don't think switch is the right construct here considering there are two cases and each tests a different variable. Why not if else?

internal/controller/proxmoxcluster_controller.go
```diff
@@ -168,6 +168,23 @@ func (r *ProxmoxClusterReconciler) reconcileNormal(ctx context.Context, clusterS
...
+		case clusterScope.ProxmoxCluster.Spec.ControlPlaneEndpoint.Host == "":
```
wikkyk, line 173:
```suggestion
		if clusterScope.ProxmoxCluster.Spec.ControlPlaneEndpoint.Host == "" {
```

internal/controller/proxmoxcluster_controller.go
```diff
@@ -168,6 +168,23 @@ func (r *ProxmoxClusterReconciler) reconcileNormal(ctx context.Context, clusterS
...
+			clusterScope.Logger.Info("ProxmoxCluster is not ready, missing or waiting for a ControlPlaneEndpoint host")
+
+			conditions.MarkFalse(clusterScope.ProxmoxCluster, infrav1alpha1.ProxmoxClusterReady, infrav1alpha1.MissingControlPlaneEndpointReason, clusterv1.ConditionSeverityWarning, "The ProxmoxCluster is missing or waiting for a ControlPlaneEndpoint host")
+
+			return ctrl.Result{Requeue: true}, nil
+		case clusterScope.ProxmoxCluster.Spec.ControlPlaneEndpoint.Port == 0:
```
wikkyk, line 179:
```suggestion
		} else if clusterScope.ProxmoxCluster.Spec.ControlPlaneEndpoint.Port == 0 {
```

wikkyk (commented):
Just this minor tidbit.

PS sorry for the late review, with summer holidays and all we have a limited time budget at the moment.

prometherion:
@wikkyk no worries, no pressure at all, I truly understand the efforts in maintaining open source, especially considering summer has approached.

I addressed the changes you requested, and noticed e2e failed tho, could you help me understand what's broken here?

wikkyk:
@prometherion the e2e fails are on us. We'll move them to a more reliable infra Soon™ :)

mcbenjemaa:
What is missing here, and what we can do?

prometherion:
@mcbenjemaa I just fixed the broken generated files, an e2e would be cool despite it seems a bit flaky according to the latest runs.

I can try also to provide a small recorded smoke test by showing the integration with Proxmox and Kamaji: unfortunately, providing a proper test is a bit complicated since the dependencies between the moving parts.

mcbenjemaa:
> @mcbenjemaa I just fixed the broken generated files, an e2e would be cool despite it seems a bit flaky according to the latest runs.
> 
> I can try also to provide a small recorded smoke test by showing the integration with Proxmox and Kamaji: unfortunately, providing a proper test is a bit complicated since the dependencies between the moving parts.

Can you share with me the manifest used to provision a proxmox cluster?

internal/controller/proxmoxcluster_controller.go
```diff
@@ -168,6 +168,22 @@ func (r *ProxmoxClusterReconciler) reconcileNormal(ctx context.Context, clusterS
 	// If the ProxmoxCluster doesn't have our finalizer, add it.
 	ctrlutil.AddFinalizer(clusterScope.ProxmoxCluster, infrav1alpha1.ClusterFinalizer)
 
+	if clusterScope.ProxmoxCluster.Spec.ExternalManagedControlPlane {
+		if clusterScope.ProxmoxCluster.Spec.ControlPlaneEndpoint.Host == "" {
```
mcbenjemaa, line 172: I think we should add a check,  `if clusterScope.ProxmoxCluster.Spec.ControlPlaneEndpoint != nil`

wikkyk: good catch - it actually panicked the test

mcbenjemaa (approved):
LGTM

wikkyk:
Finally! Thank you for your contribution! :)

prometherion:
@mcbenjemaa absolutely, it's on my todo list!

prometherion:
Sharing also here just a reference to get this working with Kamaji as externally managed Control Plane

```yaml
apiVersion: cluster.x-k8s.io/v1beta1
kind: Cluster
metadata:
  name: proxmox-quickstart
  namespace: default
spec:
  clusterNetwork:
    pods:
      cidrBlocks:
      - REDACTED/REDACTED
  controlPlaneRef:
    apiVersion: controlplane.cluster.x-k8s.io/v1alpha1
    kind: KamajiControlPlane
    name: proxmox-quickstart
  infrastructureRef:
    apiVersion: infrastructure.cluster.x-k8s.io/v1alpha1
    kind: ProxmoxCluster
    name: proxmox-quickstart
---
apiVersion: controlplane.cluster.x-k8s.io/v1alpha1
kind: KamajiControlPlane
metadata:
  name: proxmox-quickstart
  namespace: default
spec:
  dataStoreName: default
  addons:
    coreDNS: { }
    kubeProxy: { }
  kubelet:
    cgroupfs: systemd
    preferredAddressTypes:
    - InternalIP
  network:
    serviceType: LoadBalancer
  deployment:
  replicas: 2
  version: 1.29.7
---
apiVersion: infrastructure.cluster.x-k8s.io/v1alpha1
kind: ProxmoxCluster
metadata:
  name: proxmox-quickstart
  namespace: default
spec:
  allowedNodes:
  - pve
  dnsServers:
  - REDACTED
  - REDACTED
  externalManagedControlPlane: true
  ipv4Config:
    addresses:
    - REDACTED-REDACTED
    gateway: REDACTED
    prefix: REDACTED
---
apiVersion: cluster.x-k8s.io/v1beta1
kind: MachineDeployment
metadata:
  name: proxmox-quickstart-workers
  namespace: default
spec:
  clusterName: proxmox-quickstart
  replicas: 2
  selector:
    matchLabels: null
  template:
    metadata:
      labels:
        node-role.kubernetes.io/node: ""
    spec:
      bootstrap:
        configRef:
          apiVersion: bootstrap.cluster.x-k8s.io/v1beta1
          kind: KubeadmConfigTemplate
          name: proxmox-quickstart-worker
      clusterName: proxmox-quickstart
      infrastructureRef:
        apiVersion: infrastructure.cluster.x-k8s.io/v1alpha1
        kind: ProxmoxMachineTemplate
        name: proxmox-quickstart-worker
      version: v1.29.7
---
apiVersion: infrastructure.cluster.x-k8s.io/v1alpha1
kind: ProxmoxMachineTemplate
metadata:
  name: proxmox-quickstart-worker
  namespace: default
spec:
  template:
    spec:
      disks:
        bootVolume:
          disk: scsi0
          sizeGb: REDACTED
      format: qcow2
      full: true
      memoryMiB: REDACTED
      network:
        default:
          bridge: REDACTED
          model: virtio
      numCores: REDACTED
      numSockets: REDACTED
      sourceNode: pve
      templateID: REDACTED
---
apiVersion: bootstrap.cluster.x-k8s.io/v1beta1
kind: KubeadmConfigTemplate
metadata:
  name: proxmox-quickstart-worker
  namespace: default
spec:
  template:
    spec:
      joinConfiguration:
        nodeRegistration:
          kubeletExtraArgs:
            provider-id: proxmox://'{{ ds.meta_data.instance_id }}'
      users:
      - name: root
        sshAuthorizedKeys:
        - REDACTED
```

I wasn't able to let worker nodes join the Control Plane mostly because I'm working on a `kind` environment, and Proxmox VE deployed in a Virtual Machine via Vagrant.

But overall everything's look good from the `clusterctl describe` command, sic:

```
NAME                                                         READY  SEVERITY  REASON                       SINCE  MESSAGE                                                       
Cluster/proxmox-quickstart                                   True                                          7m30s                                                                 
├─ClusterInfrastructure - ProxmoxCluster/proxmox-quickstart  True                                          7m43s                                                                 
├─ControlPlane - KamajiControlPlane/proxmox-quickstart                                                                                                                           
└─Workers                                                                                                                                                                        
  └─MachineDeployment/proxmox-quickstart-workers             False  Warning   WaitingForAvailableMachines  7m43s  Minimum availability requires 2 replicas, current 0 available
    ├─Machine/proxmox-quickstart-workers-ft5tv-gfxf8         False  Error     VMProvisionFailed            35s    1 of 2 completed                                               
    └─Machine/proxmox-quickstart-workers-ft5tv-gjnmf         False  Info      PoweringOn                   5m14s  1 of 2 completed
```

PR #133, opened 2024-03-12
feat(vlan): Add vlan tag support for VM reconcile.

CiraciNicolo:
*Description of changes:*

This commit implement VLAN tag support when creating - or more in general reconciling - a ProxmoxMachine. This feature implements a new optional value in the ProxmoxMachine CRS, defined in the NetworkDevice type. The value is an integer, delimited in the range 1-4094.

*Testing performed:*

Tests have been also implemented to prevent regression, and e2e tests have been conducted creating both cluster using the new property and without - to check past compat.


internal/service/vmservice/utils_test.go
```diff
@@ -216,3 +216,45 @@ func TestShouldUpdateNetworkDevices_NoUpdate(t *testing.T) {
...
+
+	badstrings := []string{
+		"virtio=",
+		"bridge=",
+		"uuid=7dd9b137-6a3c-4661-a4fa-375075e1776b",
+		"",
+	}
+
+	for _, m := range goodstrings {
+		mtu := extractNetworkVLAN(m.test)
```
avorima, line 242: This should be called vlan. Same thing in the badstrings loop

internal/webhook/proxmoxmachine_webhook.go
```diff
@@ -110,24 +121,51 @@ func validateNetworks(machine *infrav1.ProxmoxMachine) error {
...
-		return nil
+		return fmt.Errorf("mtu must be at least 1000 or 1, but was %d", *device.MTU)
 	}
 
-	if *device.MTU > 999 {
-		return nil
+	return nil
+}
+
+func validateNetworkDeviceVLAN(device *infrav1.NetworkDevice) error {
```
avorima, line 157: This is already covered by the OpenAPI spec, so you can remove it.

avorima (commented):
SonarCloud complains about missing coverage for the webhook package. Since the changes there are actually not needed you can just revert them and the test should pass.

CiraciNicolo:
Hi,

I've refactored the code as you suggested, and pulled the latest changes from main 😊

mcbenjemaa:
Please rebase.

I will let @avorima finish the review.

avorima (approved):
Thanks!
LGTM

PR #154, opened 2024-03-25
Unmount and deletes the cloudinit iso

mcbenjemaa:
Unmount cloudinit iso after the machine become ready


mcbenjemaa:
Closes #156

PR #157, opened 2024-03-26
fix: allow FQDNs as controlPlaneEndpoint

### LINKED ISSUE #153: `ProxmoxCluster.spec.controlPlaneEndpoint.host` does not accept FQDNs
As we discovered today, `ProxmoxCluster.spec.controlPlaneEndpoint.host` does not accept FQDNs, while CAPI's `Cluster` CR accepts them and, with disabled validation on our end, it also works just fine.

We should address this issue, by, as discussed, checking if the provided value is a FQDN, if resolve it and then continue with the validation using it's returned records.

pborn-ionos:
*Issue #, if available:*
fixes #153

*Description of changes:*
Checks if the passed `controlPlaneEndpoint.host` is a resolvable FQDN, and if true, uses the IP from it for further validation.

*Testing performed:*
See chat.

### DIFF
--- a/internal/webhook/proxmoxcluster_webhook.go
+++ b/internal/webhook/proxmoxcluster_webhook.go
@@ -20,8 +20,8 @@ package webhook
 import (
 	"context"
 	"fmt"
-	"net"
 	"net/netip"
+	"regexp"
 	"strings"
 
 	infrav1 "github.com/ionos-cloud/cluster-api-provider-proxmox/api/v1alpha1"
@@ -64,7 +64,7 @@ func (*ProxmoxCluster) ValidateCreate(_ context.Context, obj runtime.Object) (wa
 		return warnings, err
 	}
 
-	if err := validateIPs(cluster); err != nil {
+	if err := validateControlPlaneEndpoint(cluster); err != nil {
 		warnings = append(warnings, fmt.Sprintf("cannot create proxmox cluster %s", cluster.GetName()))
 		return warnings, err
 	}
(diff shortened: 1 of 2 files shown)


65278 (commented):
This works, but is a bit awkward. Since we agreed that I'll be taking over, no further action is required.

A couple of remarks:
 - `isValidFQDN` is actually a test if a hostname resolves, and has nothing to do with FQDNs. There's no requirement on a hostname needing to be an FQDN to be valid.
 - You can resolve hostnames which are illegal as actual hostnames. As an example, `_acme-challenge.your.domain` is valid to resolve, but not a hostname. This is useful for text type records.
 - We're now implicitly depending on endpointHostIP not returning square brackets. I don't think this assumption holds

mcbenjemaa:
I have tested ipv6, but it doesn't work for proxmoxcluster. But the cluster was created successfully

```
cluster.cluster.x-k8s.io/test-dev created
  1 err  kind-vault-onboard kube 
Warning: cannot create proxmox cluster test-dev
The ProxmoxCluster "test-dev" is invalid: spec.controlplaneEndpoint: Invalid value: "[2a01:c23:61ce:d200::ff6:99bf:cf43]": provided endpoint address is not a valid IP or FQDN

```


also when i put a hostname in the controlplaneendpoint, it works for cluster but not for ProxmoxCluster.


those cases don't work:

- `test.example.com`
- `test-dev`
- `"[2a01:c23:61ce:d200::ff6:99bf:cf43]"`
- `"nginx.nginx.svc.cluster.local"`

mcbenjemaa:
CAPI doesn't have any validation for this field, i think you can omit validation for the host.

mcbenjemaa:
We keep validations of IPs of the ipam, and the validation of collides with endpoint ip. 
and we omit all validation to the endpoint host.
let's just keep it simple. then capi will actually through an error if the endpoint is not correct.

wikkyk:
fixes #153

mcbenjemaa:
I have raised this PR, that will make cluster to fail if machine has failed cloud-init run

65278:
This is done here. The PR is rather small, a couple of naming changes and a function to detect hostnames.
We error if our target is not an ip or a valid hostname, otherwise anything goes.

wikkyk (approved):

Issue #72, opened 2024-01-22
Add more proxmox tags to the vms


firecyberice:
**Describe the solution you'd like**
Right now there are two tags present for the vms in proxmox:
- `go-proxmox+cloud-init` (What is the value / reason of this?)
- `ip_net0_<the IP of the vm>`

It would be nice to have additional tags / labels like:
- clustername to identify the cluster of that node
- mgmt cluster name or any other reference to the capi cluster that is managing this vm / workload cluster in case there are more management clusters on that proxmox environment.

Optional:
It would be nice to add custom tags as well (at least at some point in the future)

mcbenjemaa:
`go-proxmox+cloud-init` is needed to make sure, we unmount the cloudinit iso and destroy it, before deleting the machine.

mcbenjemaa:
The unmount of cloud-init iso is done.

firecyberice:
https://github.com/ionos-cloud/cluster-api-provider-proxmox/pull/158

PR #158, opened 2024-03-27
Add custom tags to machines

mcbenjemaa:
This PR add Custom Tags to VM while preserving the current tags.


internal/service/vmservice/vm.go
```diff
@@ -225,6 +239,20 @@ func reconcileVirtualMachineConfig(ctx context.Context, machineScope *scope.Mach
...
+	// custom tags
+	if machineScope.ProxmoxMachine.Spec.Tags != nil {
+		tags := strings.Split(machineScope.VirtualMachine.VirtualMachineConfig.Tags, ";")
+		length := len(tags)
+		for _, tag := range machineScope.ProxmoxMachine.Spec.Tags {
+			if !machineScope.VirtualMachine.HasTag(tag) {
+				tags = append(tags, tag)
+			}
+		}
+		if len(tags) > length {
```
avorima, line 251: this check does not account for removal of tags. is this intended?

internal/webhook/proxmoxmachine_webhook.go
```diff
@@ -131,3 +144,14 @@ func validateNetworkDeviceMTU(device *infrav1.NetworkDevice) error {
 
 	return nil
 }
+
+func validateTags(tags []string) error {
+	re := regexp.MustCompile(`^[a-zA-Z0-9-_]*$`)
```
avorima, line 149: Should be `+` not `*` to prevent empty string. Also, have you tried `+kubebuilder:validation:Pattern`?

internal/service/vmservice/vm.go
mcbenjemaa: Yes, so we don't delete existing tags. 
The IP tag i mean.
Which means tags are also immutable as per the machine.

internal/webhook/proxmoxmachine_webhook.go
mcbenjemaa: Hmmm, 

Let me check

internal/service/vmservice/vm.go
avorima: then you need to add this to the webhook or a CRD kubebuilder annotation

internal/service/vmservice/vm_test.go
```diff
@@ -427,6 +430,27 @@ func TestReconcileVirtualMachineConfig_ApplyConfig(t *testing.T) {
 	require.EqualValues(t, task.UPID, *machineScope.ProxmoxMachine.Status.TaskRef)
 }
 
+func TestReconcileVirtualMachineConfigTags(t *testing.T) {
```
wikkyk, line 433: Can you add a test case for multiple tags and for zero/empty tags?

firecyberice:
closes #72

PR #159, opened 2024-03-31
cloudinit: fix multiple vrf device templating

65278:
*Description of changes:*
This very little templating error made it into production. It creates the same yaml key multiple times, which makes the last VRF device win in cloudinit.

*Testing performed:*
I've added testcases for multiple vrfs, so we don't get any repeats.


mcbenjemaa:
Rebase and merge

mcbenjemaa (approved):

PR #180, opened 2024-04-18
Reduce cpu/memory e2e test machines

mcbenjemaa:
Reduce cpu/memory


wikkyk (approved):

wikkyk:
skipping e2e as it wouldn't check anything

PR #196, opened 2024-04-29
Fix post run

mcbenjemaa:


mcbenjemaa:
@wikkyk This will fix the cleanup, let's watch it.

Issue #185, opened 2024-04-24
Unify DefaultNetworkDevice and AdditionalNetworkDevices


65278:
**Describe the solution you'd like**
At the moment, Network devices are arbitrarily split in two categories, default and additional. This duplicates code and serves no purpose.
If we just have an array of NetworkDevices, and require there to be at least 1, we get the same guarantees.

**Anything else you would like to add:**
 - IP addresses for the first network device are a problem, because they're coming from the cluster object. This needs some architecture.
 - This changes public API. We'll need to go v1alpha2 and provide migration patches.

mcbenjemaa:
We need to be aware of network devices names,
for instance, the first device must have the name `net0`

wikkyk:
ref  #51

wikkyk:
done! will be released with 0.8.0/v1alpha2.

Issue #177, opened 2024-04-17
cluster-api 1.7.* support


wikkyk:
https://github.com/kubernetes-sigs/cluster-api/releases/tag/v1.7.0

wikkyk:
Do we want to continue supporting capi 1.5?

wikkyk:
As mentioned in #188, capi 1.7.1 is incompatible with controller-runtime 0.8.0. The fix is in capi main, hopefully there will be a 1.7.2 release which would contain it.

wikkyk:
Also, capi 1.5 is now EOL so we should remove the column from the support table.

PR #197, opened 2024-04-30
readme: CAPI compat table for v1alpha2

wikkyk:
*Description of changes:*
As CAPI ≤1.5 are now EOL, there is no need to keep them around in the table.
We are also planning on releasing a new contract version.
Thus, it seems to me that it makes sense to start a new table for the API break.


wikkyk:
ref  #185

wikkyk:
ref #177

wikkyk:
Updated for capi 1.8.

wikkyk:
Obsoleted by  #531

PR #202, opened 2024-05-02
e2e: skip workflow on 'skip e2e' label

### LINKED ISSUE #199: Add ability to skip e2e tests
We need an automated way to skip e2e tests when we deem them unnecessary, e.g. because they only change documentation.
This is to save resources needed to run an e2e test.
I added a label to mark such PRs and we should hook it up with the action. One way to do it would be for the e2e action to simply succeed when the label is applied, otherwise to request deployment.

wikkyk:
*Issue #, if available:*
fixes #199

*Description of changes:*
1. `labeled` and `unlabeled` added to trigger types so that the label can be modified after the PR is submitted
2. e2e job runs if the PR does not have the `skip e2e` label.

*Testing performed:*
As this modifies a `pull_request_target` workflow it will only take effect once merged to main.

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -4,11 +4,12 @@ on:
   push:
     branches: ["main"]
   pull_request_target:
-    types: ["opened", "synchronize", "reopened"]
+    types: ["opened", "synchronize", "reopened", "labeled", "unlabeled"]
     branches: ["main"]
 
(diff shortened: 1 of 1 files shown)


65278 (approved):
Looks like it'll do the right thing.

wikkyk:
indeed it works as expected
image

PR #209, opened 2024-05-12
Fix ipv4config error in example cluster-class

isZumpo:
Was getting

**.spec.ipv4Config: expected map, got &{[10.10.10.10-10.10.10.20]}**

Assuming it is because of this recent [commit](https://github.com/ionos-cloud/cluster-api-provider-proxmox/commit/032869f1590c5623ede6aecce6cfb444cbe9f855) which did some changes to the cluster-class example.

Removes the **.addresses** from the patching configuration. With these changes the validation error no longer occurs on my setup.

### DIFF
--- a/templates/cluster-class.yaml
+++ b/templates/cluster-class.yaml
@@ -394,7 +394,7 @@ spec:
           - op: add
             path: /spec/template/spec/ipv4Config
             valueFrom:
-              variable: ipv4Config.addresses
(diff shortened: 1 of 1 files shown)


mcbenjemaa (commented):
@65278 please check

65278 (approved):
Yeah, I forgot to refactor that one. Sorry about that.
LGTM.

PR #234, opened 2024-06-10
fix(mockery): add missing boilerplate for mocker

CiraciNicolo:
This lightweight commits fix the mockgen phony by providing the needed file. 

I don't know if the omission is on a voluntary basis but I think this should not break the make process.


mcbenjemaa (approved):

wikkyk (changes requested):
I consciously decided against this as it kept causing issues with the actual copyright headers.
What's breaking, exactly?

CiraciNicolo:
The mockgen phony, and therefore any make with new functions implemented or signature edits.

wikkyk:
Thanks for letting us know. I would prefer to remove the line from the mockery config instead: see #235.

PR #235, opened 2024-06-11
mockery: don't use boilerplate

wikkyk:
*Issue #, if available:*
#234

*Description of changes:*
We decided against forcing one boilerplate to keep the copyright lines correct and because it's questionable whether generated files are subject to copyright.
However, we forgot to actually remove the boilerplate file from mockery config.

*Testing performed:*
`make mockgen` no longer fails.


wikkyk:
approval in #236

PR #248, opened 2024-06-26
USe cluster namespace if no namespace provided in credentials Ref

mcbenjemaa:
This PR:

Fixes an issue, if a user provided a ProxmoxCluster manifest with external creds with only the name of secret


avorima:
You should probably also document this behavior in the CRD spec.

wikkyk (dismissed):

Issue #96, opened 2024-01-29
Too many calls when deleting machine


arteonprifti:
**What steps did you take and what happened:**
During a rolling update, I noticed multiple calls for deleting the same VM in the Proxmox Task log

![image](https://github.com/ionos-cloud/cluster-api-provider-proxmox/assets/36534667/586118ec-4526-41c3-8d61-757d8fb9302f)

Is this the normal behavior, or should there be fewer calls when deleting machines?

**Environment:**

- Cluster-api-provider-proxmox version: v0.1.0
- Kubernetes version: (use `kubectl version`): v1.28.5
- OS (e.g. from `/etc/os-release`):  ubuntu-22.04

wikkyk:
@arteonprifti we'll need to check if #278 fixed this issue.

wikkyk:
We'll need to rethink how we delete VMs, then.

wikkyk:
I would really like to have this tackled for 0.8.

linuxdaemon:
I am not sure if this is related, but I have observed this behavior on start, reconfigure, resize, etc

wikkyk:
Let's test this again once 0.8.0 final is released.

PR #278, opened 2024-08-30
Check if VM exists before deletion

rybnico:
*Description of changes:*

You can't access the status of a VM after it has been deleted if you restrict the permissions of the Proxmox user or token to a pool. The API response code isn't 404 but 403 - because the non-existent VM no longer belongs to a pool and therefore the permission check fails.

This PR adds a call to `CheckID` to check if the vmid in question exists. If it doesn't exist, we can assume that the VM has already been deleted.

I added the `CheckID` function to luthermonson/go-proxmox in this PR: https://github.com/luthermonson/go-proxmox/pull/157

It's included in the 0.1.1 release of the library.

If this PR is merged, I would be happy to provide an additional PR for the docs, extending them to include instructions for a minimum-permission PVE token.

After merging this PR, the permissions that CAPI needs are

```
 acl {
    path = "/"
    propagate = false
    role_id = "SysAudit"
  }
  acl {
    path = "/nodes
    propagate = true
    role_id = "SysAudit"
  }
  acl {
    path = "/sdn/zones/localnetwork/vmbr0/330"
    propagate = false
    role_id = "PVESDNUser"
  }
  acl {
    path = "/storage/capi_poc01_block"
    propagate = false
    role_id = "PVEDatastoreUser"
  }
  acl {
    path = "/storage/capi_poc01_file"
    propagate = false
    role_id = "PVEDatastoreAdmin"
  }
  acl {
    path = "/pool/packer
    propagate = false
    role_id = "PVETemplateUser"
  }
  acl {
    path = "/pool/capi_poc01"
    propagate = false
    role_id = "PVEVMAdmin"
  }
```

*Testing performed:*
There is currently no test for the `DeleteVM` function. Writing a test including the API mock would exceed the time I'm able to invest in this PR.


pkg/proxmox/goproxmox/api_client.go
```diff
@@ -151,6 +151,17 @@ func (c *APIClient) DeleteVM(ctx context.Context, nodeName string, vmID int64) (
...
+	if err != nil {
+		return nil, fmt.Errorf("cannot get cluster")
+	}
+
+	if vmidFree, err := cluster.CheckID(ctx, int(vmID)); vmidFree {
+		return nil, fmt.Errorf("vm with id %d does not exist", vmID)
+	} else if err != nil {
+		return nil, fmt.Errorf("cannot check if vmid %d is taken: %w", vmID, err)
+	}
+
```
mcbenjemaa, line 164:
```suggestion
	if vmidFree, err := cluster.CheckID(ctx, int(vmID)); vmidFree {
		return nil, VMIDFreeErr
	} else if err != nil {
		return nil, fmt.Errorf("cannot check if vmid %d is taken: %w", vmID, err)
	}
```

mcbenjemaa: You would need to go to `reconcileDelete` and check for that error with

 ```
 if err !=nil && errors.Is(err, VMIDFreeErr) {
  return nil
 }
 else if err != nil {
   return errors.Wrap(err, "cannot delete vm")
}

mcbenjemaa (changes requested):
please check the comments. 
but also do not understand the note regarding the permissions.

rybnico:
> but also do not understand the note regarding the permissions.

If the Proxmox user/token used by the cluster-api-provider is restricted to a pool, as in this example, it's not allowed to query the status of a non-existent VM (`404` vs. `403`, see below):
```hcl
  acl {
    path = "/pool/capi_poc01"
    propagate = false
    role_id = "PVEVMAdmin"
  }
```

The call made (here)[https://github.com/ionos-cloud/cluster-api-provider-proxmox/blob/7b75c862dd90e833c6ff1e6a34a9e1a7e340f39d/pkg/proxmox/goproxmox/api_client.go#L154] fails because the Proxmox API doesn't respond with `404` (Not Found) but with `403` (Permission denied).
To check if the VM is deleted without having permissions to access all VMs, the `CheckID` function is used which doesn't need permissions on all VMs, but only checks if the VM id is free. We can then assume that the VM doesn't exist.

pkg/proxmox/goproxmox/api_client.go
rybnico: The current implementation actually works in my test setup.
The `DeleteVM` function is called from https://github.com/ionos-cloud/cluster-api-provider-proxmox/blob/7b75c862dd90e833c6ff1e6a34a9e1a7e340f39d/internal/service/vmservice/delete.go#L37
If the resulting error matches the string `does not exist` (checked in the `VMNotFound` function), the VM object is finally removed (by removing the finalizer).
I didn't want to reimplement the error handling of the existing implementation and the error `vm with id %d does not exist` perfectly reflects the result of the check. If we throw a specific error here, we should also do the same for the other cases in `DeleteVM`, shouldn't we?

mcbenjemaa: https://github.com/ionos-cloud/cluster-api-provider-proxmox/blob/7b75c862dd90e833c6ff1e6a34a9e1a7e340f39d/internal/service/vmservice/delete.go#L38-L44

Here, you would change `if VMNotFound` to `if errors.Is(err, VMIDFreeErr)`
Or you can change the function `VMNotFound`

mcbenjemaa: Perhaps it's better to check them both.

```
if VMNotFound(err) || errors.Is(err, VMIDFreeErr) { 
```

as I started to think about it, it might some race condition happens, and you know the Proxmox API is so lazy

rybnico: Ok, I created a new `VMIDFreeErr` and return/check for it now. I hope I understood your comment correctly.

Which race condition are you thinking of?

mcbenjemaa: I'm just concerned if some race conditions happen during the deletion of the machine, 
for example, if the VMID returned is free but the VM is not completely deleted, or vice versa,
we have found that those conditions might happen when engaging with the Proxmox API.

mcbenjemaa:
Let's hope this also fixes #96

mcbenjemaa (commented):
Please fix the linter

rybnico:
> Please fix the linter

😬 Fixed the two linter errors.

mcbenjemaa (approved):

PR #286, opened 2024-09-10
feat: Add ability to specify range of VM IDs to use

rybnico:
*Description of changes:*

This feature allows you to specify a `vmidRange` on a `ProxmoxCluster` object. A free ID between `vmidRange.start` and `vmidRange.end` will then be determined and used for a new VM.

We use ranges of VM IDs in Proxmox to differentiate between projects/customers/stages. I suspect many Proxmox users do this.

While developing this feature I found a [bug](https://github.com/luthermonson/go-proxmox/pull/160) in [luthermonson/go-proxmox](https://github.com/luthermonson/go-proxmox) which will be fixed in the next release. Due to the size of this PR and possible requests from your side to merge it, I have created this PR with a dependency on the current main branch of `go-proxmox`. Of course, it shouldn't be merged until the new release of `go-proxmox` is out.

What do you think of this feature and its implementation? 

*Testing performed:*

I created tests for the new features. We also tested the functionality on our Proxmox test cluster.


api/v1alpha1/proxmoxcluster_types.go
```diff
@@ -79,6 +87,30 @@ type ProxmoxClusterSpec struct {
...
+
+// VMIDRange defines the range of VMIDs to use for VMs.
+type VMIDRange struct {
+	// VMIDRangeStart is the start of the VMID range to use for VMs.
+	// +kubebuilder:validation:Minimum=100
+	// +kubebuilder:validation:ExclusiveMinimum=false
+	// +kubebuilder:validation:Maximum=999999999
+	// +kubebuilder:validation:ExclusiveMaximum=false
+	// +kubebuilder:validation:Required
+	Start int64 `json:"start"`
```
mcbenjemaa, line 104: would it be possible to just specify start or end only?

internal/service/vmservice/vm.go
```diff
@@ -385,6 +393,28 @@ func createVM(ctx context.Context, scope *scope.MachineScope) (proxmox.VMCloneRe
 	return res, scope.InfraCluster.PatchObject()
 }
 
+func getVMID(ctx context.Context, scope *scope.MachineScope) (int64, error) {
+	if scope.InfraCluster.ProxmoxCluster.Spec.VMIDRange != nil {
+		vmidRangeStart := scope.InfraCluster.ProxmoxCluster.Spec.VMIDRange.Start
+		vmidRangeEnd := scope.InfraCluster.ProxmoxCluster.Spec.VMIDRange.End
+		if vmidRangeStart != 0 && vmidRangeEnd != 0 {
+			// Get next free vmid from the range
+			for i := vmidRangeStart; i <= vmidRangeEnd; i++ {
```
mcbenjemaa, line 402: given that we have set a range of 100 or 200, the logic here will continuously loop until it finds a free ID,
Do you think we can optimize this?

api/v1alpha1/proxmoxcluster_types.go
rybnico: Do you think this would be useful? Considering that we have to check all IDs in the range in the worst case, I wanted to limit the range.

internal/service/vmservice/vm.go
rybnico: Since we can't query for all IDs used if we don't have permissions to do so (e.g. our PVE user is limited to a pool), I don't see any other solution. I couldn't find a better PVE API endpoint for this.
It would be possible to ignore all VM IDs we currently know (all VM IDs of existing `ProxmoxMachines`), but I couldn't find a way to get all `ProxmoxMachines` from `scope`.

api/v1alpha1/proxmoxcluster_types.go
mcbenjemaa: Yeah, if we find a use case. 
We can try it.

internal/service/vmservice/vm.go
mcbenjemaa: How about we add a resource pool
For each cluster. 
And based on that pool. We will figure out the next-id.
And of course we will check that id if free or we try to retrieve new one between, start-end. 
This will limit the calls to the API. 


Or you can also try to call API, Cluster resources. 
But I'm afraid if some Proxmox datacenters has a lot of resources. 

This is just my thoughts.
Cause im just concerned about loop and calling API each time. 

Let me know what do you think?

rybnico: What do you mean by 'resource pool'? Currently the range you specify in the `ProxmoxCluster` object limits the number of calls to the endpoint.
We can't get a list of all Proxmox resources if the user's permissions are limited, so we can't know if a vmid is free or not without calling the endpoint.

Am I missing something?

api/v1alpha1/proxmoxcluster_types.go
rybnico: If we only specify `vmidRange.end`, we would have to loop from `100` (first valid ID) to `vmidRange.end`, which could cause a lot of calls.
If we only specify `vmidRange.start`, we would have to loop from `vmidRange.start` to `999999999`, which seems to be the maximum allowed vmid, which would also cause a lot of calls.
I think it's better to limit the range from the beginning by requiring `start` and `end`, what do you think?

mcbenjemaa: I agree with start-end

internal/service/vmservice/vm.go
mcbenjemaa: Let me involve @65278 in this matter,
Do you think it would be okay to do a loop in which we call the Proxmox API to check if the ID is free in each loop?

I just have some concerns if a cluster already has many machines, and this will probably get the reconciliation loop stuck until it finds a new ID, this is definitely not a good practice for the controller perspective.

mcbenjemaa: yeah, it might be a good option before checking the API.
but before the loop, you should get the list of all ProxmoxMachines so you can check if the VMID exists or not,

mcbenjemaa: there's probably a point we need to take it carefully: 
if the cluster uses `spec.credentialsRef` then you need to list only the machines related to this Proxmox Datacenter,
if not list the machines that use the default datacenter.

it's quite complex but duable.

rybnico: I moved the function `listProxmoxMachinesForCluster` from `internal/controller/proxmoxcluster_controller.go` to `pkg/scope/cluster.go` and made it public. I use this function to get all the ProxmoxMachines that belong to the Cluster. Since `spec.credentialsRef` is specified on a Cluster/ProxmoxCluster, this should only get the ProxmoxMachines that belong to one Proxmox Datacenter.

Could you please take a look and tell me if you like my implementation. Thanks in advance!

internal/service/vmservice/vm.go
```diff
@@ -385,6 +394,39 @@ func createVM(ctx context.Context, scope *scope.MachineScope) (proxmox.VMCloneRe
...
+		vmidRangeEnd := scope.InfraCluster.ProxmoxCluster.Spec.VMIDRange.End
+		if vmidRangeStart != 0 && vmidRangeEnd != 0 {
+			// Get all used vmids from existing ProxmoxMachines
+			usedVMIDs := []int64{}
+			proxmoxMachines, err := scope.InfraCluster.ListProxmoxMachinesForCluster(ctx)
+			if err != nil {
+				return 0, err
+			}
+			for _, proxmoxMachine := range proxmoxMachines {
+				usedVMIDs = append(usedVMIDs, proxmoxMachine.GetVirtualMachineID())
```
mcbenjemaa, line 409: I would add a check
```
if proxmoxMachine.GetVirtualMachineID() != -1 {
usedVMIDs = append(usedVMIDs, proxmoxMachine.GetVirtualMachineID())
}

File: internal/service/vmservice/vm.go [hunk 4]
```diff
@@ -311,10 +315,15 @@ func getMachineAddresses(scope *scope.MachineScope) ([]clusterv1.MachineAddress,
 }
 
 func createVM(ctx context.Context, scope *scope.MachineScope) (proxmox.VMCloneResponse, error) {
+	vmid, err := getVMID(ctx, scope)
+	if err != nil {
+		return proxmox.VMCloneResponse{}, err
```
mcbenjemaa, line 320: there's only one remaining thing:
you will probably need to fail the machine provisioning here.

```go
if err != nil {
  if errors.Is(err, ErrNoVMIDInRangeFree) {
    // fail the machine, by setting failureReason, and failureMessage
  }
}

```

65278 (changes requested):
Disregard the nitpicks in the review.

My problem is with the API itself:
Putting the VMID range into cluster id works for your usecase, but other people may have different use cases.
The only sensible place where to put VM IDs is into the ProxmoxMachineTemplates, as people may want to set VM ID ranges per machine kind.

Of course this means that if you want to reuse ProxmoxMachineTemplates but not VM IDs, you can't, but that's a small price to pay.

rybnico:
> My problem is with the API itself: Putting the VMID range into cluster id works for your usecase, but other people may have different use cases. The only sensible place where to put VM IDs is into the ProxmoxMachineTemplates, as people may want to set VM ID ranges per machine kind.

Very good point. I tested the functionality by specifying different vmidRanges for my control plane and worker nodes and liked it immediately.
I moved spec.vmidRange from ProxmoxCluster to ProxmoxMachine (which is used in ProxmoxMachineTemplate) and updated the tests accordingly.

mcbenjemaa:
@rybnico what is missing here!

rybnico:
> @rybnico what is missing here!

From my point of view, everything is done. @65278 requested a change [here](https://github.com/ionos-cloud/cluster-api-provider-proxmox/pull/286#pullrequestreview-2324931288), I'm still waiting for feedback.

pkg/scope/cluster.go
```diff
@@ -195,6 +195,20 @@ func (s *ClusterScope) PatchObject() error {
 	return s.patchHelper.Patch(context.TODO(), s.ProxmoxCluster)
 }
 
+// ListProxmoxMachinesForCluster returns all the ProxmoxMachines that belong to this cluster.
+func (s *ClusterScope) ListProxmoxMachinesForCluster(ctx context.Context) ([]infrav1alpha1.ProxmoxMachine, error) {
```
mcbenjemaa, line 199: you can add a test case for this function?

internal/service/vmservice/vm_test.go
```diff
@@ -181,6 +184,104 @@ func TestEnsureVirtualMachine_CreateVM_SelectNode_InsufficientMemory(t *testing.
 	require.True(t, machineScope.HasFailed())
 }
 
+func TestEnsureVirtualMachine_CreateVM_VMIDRange(t *testing.T) {
```
mcbenjemaa, line 187: can you add a case here where the ProxmoxMachine VMID = -1

mcbenjemaa (commented):
thanks for your contribution, 
please check the added comments for 2 minor things, and please rebase.
I will make sure to get this merged soon)

api/v1alpha1/proxmoxmachine_types_test.go
```diff
@@ -286,4 +286,35 @@ var _ = Describe("ProxmoxMachine Test", func() {
 			Expect(k8sClient.Create(context.Background(), dm)).Should(MatchError(ContainSubstring("should be less than or equal to 4094")))
 		})
 	})
+
+	Context("VMIDRange", func() {
+		It("Should not allow only allow spec.vmidRange.start >= 100", func() {
```
wikkyk, line 291: please fix the wording here

api/v1alpha1/proxmoxmachine_types_test.go
```diff
@@ -286,4 +286,35 @@ var _ = Describe("ProxmoxMachine Test", func() {
...
+			dm := defaultMachine()
+			dm.Spec.VMIDRange = &VMIDRange{
+				Start: 1,
+			}
+			Expect(k8sClient.Create(context.Background(), dm)).Should(MatchError(ContainSubstring("should be greater than or equal to 100")))
+		})
+		It("Should not allow only allow spec.vmidRange.end >= 100", func() {
```
wikkyk, line 298: ditto

api/v1alpha1/proxmoxmachine_types_test.go
```diff
@@ -286,4 +286,35 @@ var _ = Describe("ProxmoxMachine Test", func() {
...
+			dm := defaultMachine()
+			dm.Spec.VMIDRange = &VMIDRange{
+				End: 1,
+			}
+			Expect(k8sClient.Create(context.Background(), dm)).Should(MatchError(ContainSubstring("should be greater than or equal to 100")))
+		})
+		It("Should not allow only allow spec.vmidRange.start if spec.vmidRange.end is set", func() {
```
wikkyk, line 305: ditto

api/v1alpha1/proxmoxmachine_types_test.go
```diff
@@ -286,4 +286,35 @@ var _ = Describe("ProxmoxMachine Test", func() {
...
+			dm := defaultMachine()
+			dm.Spec.VMIDRange = &VMIDRange{
+				Start: 100,
+			}
+			Expect(k8sClient.Create(context.Background(), dm)).Should(MatchError(ContainSubstring("spec.vmidRange.end in body should be greater than or equal to 100")))
+		})
+		It("Should not allow only allow spec.vmidRange.end if spec.vmidRange.start is set", func() {
```
wikkyk, line 312: ditto

wikkyk (changes requested):
Minor tidbits from me.

go.mod
```diff
@@ -10,7 +10,7 @@ require (
 	github.com/go-logr/logr v1.4.2
 	github.com/google/uuid v1.6.0
 	github.com/jarcoal/httpmock v1.3.1
-	github.com/luthermonson/go-proxmox v0.1.1
+	github.com/luthermonson/go-proxmox v0.2.1
```
wikkyk, line 13: This should be done in a separate PR. If you really need it the bump, wait until #306 is merged and rebase.

wikkyk (changes requested):

go.mod
mcbenjemaa: I don't mind if the bump is here,
I don't see the value of waiting, 
he just bumped to add the new feature.

wikkyk: #306 is merged

mcbenjemaa (commented):
I tested this, but the machines are not getting ready:

```
I1105 12:54:10.082776   51298 proxmoxmachine_controller.go:216] "VM state is not reconciled" controller="proxmoxmachine" controllerGroup="infrastructure.cluster.x-k8s.io" controllerKind="ProxmoxMachine" ProxmoxMachine="default/test-1-control-plane-qjts8" namespace="default" name="test-1-control-plane-qjts8" reconcileID="50e2ed5f-19f0-4a06-9615-7b9abc9290c8" machine="default/test-1-control-plane-qjts8" cluster="default/test-1" expectedVMState="ready" actualVMState="pending"
```

did you tested this?

rybnico:
> However, I noticed the Control plane machines starts from `start+1`

I tested again with the following settings:

Control plane:
```yaml
      vmIDRange:
        end: 1009
        start: 1000
```

Worker nodes:
```yaml
      vmIDRange:
        end: 1099
        start: 1010
```

The whole range was free when I ran `clusterctl alpha rollout restart kubeadmcontrolplane/poc01-control-plane ` and `clusterctl alpha rollout restart machinedeployment/poc01-workers`.

This is the result after the rollout has finished:
image

I'm not sure what happened on your end. Perhaps the start VM ID was already taken?

mcbenjemaa:
> > However, I noticed the Control plane machines starts from `start+1`
> 
> I tested again with the following settings:
> 
> Control plane:
> 
> ```yaml
>       vmIDRange:
>         end: 1009
>         start: 1000
> ```
> 
> Worker nodes:
> 
> ```yaml
>       vmIDRange:
>         end: 1099
>         start: 1010
> ```
> 
> The whole range was free when I ran `clusterctl alpha rollout restart kubeadmcontrolplane/poc01-control-plane ` and `clusterctl alpha rollout restart machinedeployment/poc01-workers`.
> 
> This is the result after the rollout has finished: image
> 
> I'm not sure what happened on your end. Perhaps the start VM ID was already taken?

after your last commit the `vmidRange` doesn't even work now,

rybnico:
> after your last commit the `vmidRange` doesn't even work now,

I have renamed all occurrences of `vmidRange` to `vmIDRange` - did you adjust the CRDs and CRs before your tests?

mcbenjemaa:
that's what i got.

```
      vmIDRange:
        start: 1000
        end: 1100
```

```
      vmIDRange:
        start: 2000
        end: 2100
```

Screenshot 2024-11-06 at 12 18 25

rybnico:
> that's what i got.

I cannot reproduce the issue. When I create a new VM, the controller reliably picks the first free VM ID from the `vmIDRange`, on both control plane and worker nodes.

Can you please have a look in the debug log at the API call to `/api2/json/cluster/nextid` that is made to determine the next free VM ID?

It should look something like this and it happens just before the VM clone task is started:
```
I1106 23:20:05.653862   68929 find.go:63] "vmid doesn't exist yet" controller="proxmoxmachine" controllerGroup="infrastructure.cluster.x-k8s.io" controllerKind="ProxmoxMachine" ProxmoxMachine="default/poc01-control-plane-6zp2l" namespace="default" name="poc01-control-plane-6zp2l" reconcileID="7c8f9f1f-0
38f-4bcc-ad63-e513ddb96a67" machine="default/poc01-control-plane-6zp2l" cluster="default/poc01"
I1106 23:20:05.653904   68929 logger.go:52] SEND: GET - https://proxmoxtest01:8006/api2/json/cluster/status
I1106 23:20:05.734977   68929 logger.go:52] RECV: 200 - 200 OK
I1106 23:20:05.734997   68929 logger.go:52] BODY: {"data":[{"level":"","type":"node","name":"proxmoxtest01","local":1,"nodeid":0,"online":1,"id":"node/proxmoxtest01","ip":"10.1.103.50"}]}
I1106 23:20:05.735016   68929 logger.go:52] SEND: GET - https://proxmoxtest01:8006/api2/json/cluster/nextid?vmid=1000
I1106 23:20:05.750891   68929 logger.go:52] RECV: 200 - 200 OK
I1106 23:20:05.750898   68929 logger.go:52] BODY: {"data":"1000"}
```

rybnico:
@mcbenjemaa Have you had a chance to test this again with debug output enabled? Is there anything I can do to help you test this branch?

mcbenjemaa:
I will test it again and let you know

mcbenjemaa:
> I will test it again and let you know

Testing with VMIDRanges is okay.
But, without ranges, there's something strange!

mcbenjemaa:
> > Testing with VMIDRanges is okay. But, without ranges, there's something strange!
> 
> Ok, I will also test without VMIDRanges. Can you be a bit more specific about "something strange"?

The machines are failing and are a bit slow. Honestly, I don't know whether it is because of the code or because some issues are happening in my setup.
let's run e2e tests again.

internal/service/vmservice/vm.go
```diff
@@ -319,10 +323,15 @@ func getMachineAddresses(scope *scope.MachineScope) ([]clusterv1.MachineAddress,
 }
 
 func createVM(ctx context.Context, scope *scope.MachineScope) (proxmox.VMCloneResponse, error) {
+	vmid, err := getVMID(ctx, scope)
+	if err != nil {
```
mcbenjemaa, line 327: check my comment https://github.com/ionos-cloud/cluster-api-provider-proxmox/pull/286/files#r1768603393

rybnico:
@mcbenjemaa I can't see from the logs why the e2e tests failed.

The logs from the Proxmox cluster would be interesting, though:
> STEP: Dumping logs from the "capmox-e2e-rzk79z" workload cluster @ 12/02/24 14:35:32.973
Unable to get logs for workload Cluster create-workload-cluster-1hh797/capmox-e2e-rzk79z: log collector is nil.

Any idea why it failed?

internal/service/vmservice/vm.go
```diff
@@ -393,6 +406,41 @@ func createVM(ctx context.Context, scope *scope.MachineScope) (proxmox.VMCloneRe
...
+			proxmoxMachines, err := scope.InfraCluster.ListProxmoxMachinesForCluster(ctx)
+			if err != nil {
+				return 0, err
+			}
+			for _, proxmoxMachine := range proxmoxMachines {
+				if proxmoxMachine.GetVirtualMachineID() != -1 {
+					usedVMIDs = append(usedVMIDs, proxmoxMachine.GetVirtualMachineID())
+				}
+			}
+			// Get next free vmid from the range
```
mcbenjemaa, line 425: @rybnico can you just split the getVMID function, so that sonarcloud don't complain?

mcbenjemaa: you can move this section, `// Get next free vmid from the range` to its own function

mcbenjemaa (approved):
Thanks for your contribution.

LGTM

65278 (commented):
The PR looks good to me (I don't even have a nitpick). I'll give it a try and then greenlight.

65278 (approved):

wikkyk (approved):

PR #290, opened 2024-09-22
Support toggling provisioning checks

### LINKED ISSUE #183: Support Kubernetes v1.30
https://kubernetes.io/blog/2024/04/17/kubernetes-v1-30-release/

ekarlso:
cf #183

*Testing performed:*
Added / adapted tests


api/v1alpha1/proxmoxmachine_types.go
```diff
@@ -50,6 +50,13 @@ const (
 	IPV6Format = "v6"
 )
 
+type ProxmoxMachineChecks struct {
+	// Skip checking CloudInit which can be very useful for specific Operating Systems like TalOS
+	// +optional
+	SkipCloudInitStatus *bool `json:"skipCloudInitStatus,omitempty"`
+	SkipQemuGuestAgent  *bool `json:"skipQemuGuestAgent,omitempty"`
```
mcbenjemaa, line 57: Each field should has it own, description and markers

internal/service/vmservice/vm.go
```diff
@@ -108,22 +108,47 @@ func ReconcileVM(ctx context.Context, scope *scope.MachineScope) (infrav1alpha1.
 	return vm, nil
 }
 
+func skipQemuGuestCheck(mc *scope.MachineScope) bool {
+	if mc.ProxmoxMachine.Spec.Checks != nil {
+		return ptr.Equal(ptr.To(true), mc.ProxmoxMachine.Spec.Checks.SkipQemuGuestAgent)
```
mcbenjemaa, line 113: Use ptr.Deref

wikkyk:
@ekarlso I edited your comment to remove the word 'fixes' as otherwise merging this PR would close the issue prematurely.

pkg/proxmox/goproxmox/api_client.go
```diff
@@ -300,3 +300,11 @@ func (c *APIClient) CloudInitStatus(ctx context.Context, vm *proxmox.VirtualMach
 
 	return false, nil
 }
+
+func (c *APIClient) QemuAgentStatus(ctx context.Context, vm *proxmox.VirtualMachine) error {
```
mcbenjemaa, line 304: You should have called this function in `CloudInitStatus()`

internal/service/vmservice/vm.go
```diff
@@ -108,22 +108,46 @@ func ReconcileVM(ctx context.Context, scope *scope.MachineScope) (infrav1alpha1.
 	return vm, nil
 }
 
+func skipQemuGuestCheck(mc *scope.MachineScope) bool {
```
mcbenjemaa, line 111: please move those functions to the machineScope

internal/service/vmservice/vm.go
```diff
@@ -108,22 +108,46 @@ func ReconcileVM(ctx context.Context, scope *scope.MachineScope) (infrav1alpha1.
...
+	if mc.ProxmoxMachine.Spec.Checks != nil {
+		return ptr.Deref(mc.ProxmoxMachine.Spec.Checks.SkipQemuGuestAgent, false)
+	}
+
+	return false
+}
+
+func skipCloudInitCheck(mc *scope.MachineScope) bool {
```
mcbenjemaa, line 119: same:
please move those functions to the machineScope

internal/service/vmservice/vm.go
```diff
@@ -108,22 +108,46 @@ func ReconcileVM(ctx context.Context, scope *scope.MachineScope) (infrav1alpha1.
...
+			return true, errors.Wrap(err, "error waiting for agent")
 		}
-		if errors.Is(goproxmox.ErrCloudInitFailed, err) {
-			conditions.MarkFalse(machineScope.ProxmoxMachine, infrav1alpha1.VMProvisionedCondition, infrav1alpha1.VMProvisionFailedReason, clusterv1.ConditionSeverityError, err.Error())
-			machineScope.SetFailureMessage(err)
-			machineScope.SetFailureReason(capierrors.MachineStatusError("BootstrapFailed"))
+	}
+
+	if !skipCloudInitCheck(machineScope) {
+		if running, err := machineScope.InfraCluster.ProxmoxClient.CloudInitStatus(ctx, machineScope.VirtualMachine); err != nil || running {
```
mcbenjemaa, line 140: If you set `skipQemuGuestCheck` this will call the qemuGuestCheck nevertheless, please check my comment in the api_client.

Here you need to check if  `!skipQemuGuestCheck && !skipCloudInitCheck ` cause you can't get the cloud-init status unless you have qemu-guest-agent

mcbenjemaa (changes requested):

mcbenjemaa:
There is still failing tests

mcbenjemaa (approved):

ricottatosta:
Dear IONOS Team,
I believe this is not the right way to address all the issues around cloud-init and Talos. And the way it will be addressed also involves flatcar support in some way. 
Talos image can get qemu-agent via custom image, so it needn't to check it. But if there is any other issue regarding the agent, it can be addressed.
The problem involving cloud-init and Talos is related to cloud-init datasource. Up to date, any os running in a vm must expect network-config to be in cloud-init netplan format and must install cloud-init binary.
On the other hand, Talos expects network-config to be in cloud-init nocloud format and has no cloud-init binary installed.
Flatcar has another config file format, in json if I'm not wrong.
For all this reasons, I think all these issues should be addressed more schematically, maybe using a schema like this in proxmox_machine_type.go:

spec:
    config:
        engine: cloud-init|flatcar
        flatcar: {}
        cloudInit:
            dataSource: netplan|nocloud
    qemuAgent: true|false

With this new arrangement in place, it should be possible to address all the issues around that involve both reconciliation activities and engine config format in general.
I have started forking your project in order to be able to create a Pull Request. If you are interested in my idea, I will try to arrange the code. But be aware, I'm less than a newbie at GoLang. It will need a review.
I hope you will change your mind.

mcbenjemaa:
Hey your Proposal is nice
but we need to keep things in mind,
in debian-based os, we check cloud-init status via the qemuAgent. 
so we need a way of skipping such a check.
in flatcar we can run qemuAgent but not cloud-init.

```yaml
spec:
  config:
    engine: cloud-init|flatcar
    flatcar: {}
    cloudInit:
      dataSource: netplan|nocloud
  qemuAgent: true|false

```
here is some of my notes:

- Having `qemuAgent` field is great
- `config.engine` is also nice. however we can also check that based on generated bootstrap data.
- `config.cloudInit.dataSource` I don't know why you need such a field, please explain.

However, can you also explain how we can actually implement the cloud-init checks and Qemu and under which conditions?

Finally, a word of advice, please create a new issue and start a discussion around the API additions or changes in the issue, before starting the PR,
you need the API approved, so you don't change it when you start the PR.
we would appreciate your contributions.

mcbenjemaa:
@ekarlso 
Is that the case?
I thought the network-config v2 works well in your setup.


@ricottatosta Open new issue, and let's discuss there.

PR #292, opened 2024-09-30
cloud-init: Add provider-id to metadata

ekarlso:
Some operating system(s) like Talos use provider-id for their
cloud-provider implementation in the format of <provider>://<id>
to be picked up by the cloud-init impl. Having it available is pretty standard.

*Issue #, if available:*

*Description of changes:*

*Testing performed:*


pkg/cloudinit/metadata.go
```diff
@@ -18,7 +18,10 @@ package cloudinit
 
 const (
 	metadataTPl = `instance-id: {{ .InstanceID }}
-local-hostname: {{ .Hostname }}`
+local-hostname: {{ .Hostname }}
+hostname: {{ .Hostname }}
```
mcbenjemaa, line 22: hostname is not a metadata it's userdata

pkg/cloudinit/metadata.go
```diff
@@ -18,7 +18,10 @@ package cloudinit
...
+provider-id: {{ .ProviderID }}
```
mcbenjemaa, line 23: you can use `provider-id: {{ .InstanceID }}`

pkg/cloudinit/types.go
```diff
@@ -21,6 +21,7 @@ package cloudinit
 type BaseCloudInitData struct {
 	Hostname          string
 	InstanceID        string
+	ProviderID        string
```
mcbenjemaa, line 24: and get rid of this)

mcbenjemaa (approved):

mcbenjemaa (approved):

PR #314, opened 2024-10-29
Skip e2e based on label

mcbenjemaa:
The idea is to read a label  with prefix `skipe-e2e/*`what comes after the prefix will be skipped. 
There are 3 Labels for now: 

- `skip-e2e/Flatcar` will skip tests labeled with 'Flatcar'
- `skip-e2e/Generic` will skip tests labeled with 'Generic'
- `skip-e2e/none` will not skip any tests

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -32,8 +32,36 @@ jobs:
         with:
           go-version-file: go.mod
 
+      - name: Get e2e skip label
+        id: get_labels
+        run: |
+          # Look for label that starts with skip-e2e/
+          skip_label="Flatcar"
+          
+          # Convert labels to JSON and check if it is not null
+          labels_json='${{ toJson(github.event.pull_request.labels) }}'
+          
+          if [[ "$labels_json" != "null" ]] && [[ "$labels_json" != "[]" ]]; then
+              # Use jq to parse the labels and extract names
+              labels=$(echo $labels_json | jq -r '.[].name')
+          
+              # Iterate through the names and check for the prefix "skip-e2e/"
+              for label_name in $labels; do
+                if [[ $label_name == skip-e2e/* ]]; then
+                  skip_label="${label_name#skip-e2e/}" # Remove the prefix to get the value
+                  break
+                fi
+              done
+          else
+              echo "No labels found."
+          fi
+          
+          echo "SKIP_E2E=$skip_label"
+          # Output the label value, or set it to "none" if not found
+          echo "SKIP_E2E=$skip_label" >> "$GITHUB_ENV"
+
       - name: Run e2e tests
-        run: "make test-e2e"
+        run: "make test-e2e GINKGO_SKIP=${{ env.SKIP_E2E }}"
 
       - name: Cleanup kind clusters
         uses: gacts/run-and-post-run@v1
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -16,7 +16,6 @@ jobs:
       PROXMOX_URL: ${{ secrets.PROXMOX_URL }}
       PROXMOX_TOKEN: ${{ secrets.PROXMOX_TOKEN }}
       PROXMOX_SECRET: ${{ secrets.PROXMOX_SECRET }}
-      SKIP_E2E: "Flatcar"
     steps:
       - name: Check out branch ${{ github.ref }}
         if: ${{ github.event_name == 'push' }}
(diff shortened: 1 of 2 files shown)


mcbenjemaa:
This has been tested, 
If we add `skip-e2e/Flatcar` or `skip-e2e/Generic` it will skip those test cases.

wikkyk:
What's the use case for `skip-e2e/none`?

mcbenjemaa:
by default, the test cases will skip `Flatcar`.
if you add `skip-e2e/none` will run all test cases.

.github/workflows/e2e.yml
```diff
@@ -32,8 +32,36 @@ jobs:
         with:
           go-version-file: go.mod
 
+      - name: Get e2e skip label
+        id: get_labels
+        run: |
+          # Look for label that starts with skip-e2e/
+          skip_label="Flatcar"
```
mcbenjemaa, line 39: default skip

wikkyk:
Those labels have too many negatives, making it difficult to understand what they do. How about this:

| label  | behaviour |
| ------------- | ------------- |
| none (default) | run `Generic` tests  |
| https://github.com/ionos-cloud/cluster-api-provider-proxmox/labels/skip%20e2e (could be renamed to `e2e/skip` or `e2e/none`)| do not run any e2e tests |
| `e2e/flatcar` | run `Generic` and `Flatcar` tests |

IMHO the meaning is clearer, we aren't don't not disabling tests. We are also keeping the current behaviour of defaulting to `Generic` e2e tests, with `Flatcar` tests or no tests opt-in. `e2e/flatcar` would override `e2e/none` if both are set.

mcbenjemaa:
In theory, we can add a custom label in the PR to also skip multiple test cases, e.g. `skip-e2e/Generic|Flatcar`
which will skip all cases with Flatcar and Generic

wikkyk:
How about this? We will need to change the labels but imho the intent is clearer:

| Label | Behaviour |
|---|---|
| none | Run `Generic` tests only |
| `e2e/none` | Do not run any tests |
| `e2e/flatcar` | Run `Generic` + `Flatcar` tests |

The code isn't quite as generic but it only uses action logic to figure out `GINKGO_SKIP`.

.github/workflows/e2e.yml
```diff
@@ -32,8 +32,16 @@ jobs:
         with:
           go-version-file: go.mod
 
+      - name: Check if requested to skip Flatcar tests
```
mcbenjemaa, line 35: this means, you need to add a step for each $feature

.github/workflows/e2e.yml
```diff
@@ -32,8 +32,16 @@ jobs:
...
+        if: ${{ ! contains( github.event.pull_request.labels.*.name, 'e2e/flatcar') }}
+        run: echo 'SKIP_E2E="Flatcar"' >> "$GITHUB_ENV"
+
+      - name: Check if requested to skip Generic tests
+        if: ${{ contains( github.event.pull_request.labels.*.name, 'e2e/skip-generic') }}
+        run: echo "SKIP_E2E='$SKIP_E2E Generic'" >> "$GITHUB_ENV"
```
mcbenjemaa, line 41: the delimiter should be `|`

mcbenjemaa: the script here would need to check if anything defined in the variable `SKIP_E2E` so add the delimiter `|` or just write `Generic`

mcbenjemaa:
The action `.github/workflows/e2e-comment.yml` I added:
Will add a new comment to each new PR created.

the comment will look like this:

### 🚀  e2e tests  run
  We add labels to the PRs to control the e2e test runs by running specific tests and skipping some test contexts,
  please follow this guide:
  
  | Label | Behaviour |
  |--|--|
  | none | Run `Generic` tests |
  | https://github.com/ionos-cloud/cluster-api-provider-proxmox/labels/e2e%2Fnone | Do not run any tests (overrides all e2e/ flags) |
  | https://github.com/ionos-cloud/cluster-api-provider-proxmox/labels/e2e%2Fflatcar | Add `Flatcar` tests to the run |
  | https://github.com/ionos-cloud/cluster-api-provider-proxmox/labels/e2e%2Fskip-generic | Remove `Generic` tests from the run |
  
  ℹ️ Ask a `Member` to add the requested labels if you don't have enough permissions.

wikkyk:
On second thought, I'm not so sure I agree with the ability to skip the generic tests. The whole point of e2e is to have an integration test to catch issues introduced accidentally through interactions between interdependent components. As such, the basic test should *always* run (except for docs etc), even if a change seems to only affect one area.

I also don't think we need the comment, only maintainers can ok an e2e run and it is their job to verify the labels. It's the reviewer's, not the submitter's remit to choose the tests because it is the reviewer who has to be satisfied the code can be merged.

wikkyk:
Aye, I think `e2e/none` and `e2e/flatcar` is the way to go. If there are any additional optional tests in the future, they can be implemented like flatcar.

mcbenjemaa:
sure)

.github/workflows/e2e.yml
mcbenjemaa: can you adjust the file then?

mcbenjemaa:
LGTM 
You can Approve the PR.

wikkyk:
I'll merge and change the label.

wikkyk (approved):

Issue #303, opened 2024-10-18
add vlan to 'clusterctl generate'


lknite:
**Describe the solution you'd like**

vlan works when added to a network configuration in a cluster.yaml to be deployed via cluster-api, however, when creating a cluster using 'clusterctl generate' the 'VLAN' value does not make it into the cluster.yaml file.

Add 'VLAN' as an optional value to ~/.cluster-api/clusterctl.yaml & let it be used when running 'clusterctl generate'.

e.g.:
```
## -- networking configuration-- ##
CONTROL_PLANE_ENDPOINT_IP: "10.4.0.2"                         # The IP that kube-vip is going to use as a control plane endpoint
NODE_IP_RANGES: "[10.4.0.20-10.4.0.254]"                      # The IP ranges for Cluster nodes
GATEWAY: "10.4.0.1"                                           # The gateway for the machines network-config.
IP_PREFIX: "24"                                               # Subnet Mask in CIDR notation for your node IP ranges
DNS_SERVERS: "[10.0.0.2,10.0.0.3]"                            # The dns nameservers for the machines network-config.
BRIDGE: "vmbr0"                                               # The network bridge device for Proxmox VE VMs
VLAN: 1400                                                    # (Optional) VLAN Tag
```

PR #329, opened 2024-11-19
Add support for setting VLAN in network devices through config

Luzifer:
*Issue #, if available:* #303

*Description of changes:*

- Add `VLAN`, `SECONDARY_VLAN` and `EXT_SERVICE_VLAN` variables to the template to configure VLAN on the network devices

*Testing performed:*

- Tested generation of template with and without `VLAN` set: Expected behavior, without ENV-var a `kubectl describe` shows no configured `vlan`, with `VLAN` set to `2` it shows `vlan: 2`.

### DIFF
--- a/docs/Usage.md
+++ b/docs/Usage.md
@@ -91,6 +91,7 @@ GATEWAY: "10.10.10.1"                                         # The gateway for 
 IP_PREFIX: "25"                                               # Subnet Mask in CIDR notation for your node IP ranges
 DNS_SERVERS: "[8.8.8.8,8.8.4.4]"                              # The dns nameservers for the machines network-config.
 BRIDGE: "vmbr1"                                               # The network bridge device for Proxmox VE VMs
+VLAN: "2"                                                     # (optional) The VLAN to put the bridge device into
 
 ## -- xl nodes -- ##
 BOOT_VOLUME_DEVICE: "scsi0"                                   # The device used for the boot disk.
--- a/templates/cluster-class-calico.yaml
+++ b/templates/cluster-class-calico.yaml
@@ -982,6 +982,7 @@ spec:
         default:
           bridge: ${BRIDGE}
           model: virtio
+          vlan: ${VLAN:=null}
 ---
 kind: ProxmoxMachineTemplate
 apiVersion: infrastructure.cluster.x-k8s.io/v1alpha1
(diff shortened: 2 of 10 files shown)


mcbenjemaa (approved):
LGTM

mcbenjemaa:
anything missing?

docs/Usage.md
```diff
@@ -91,6 +91,7 @@ GATEWAY: "10.10.10.1"                                         # The gateway for
 IP_PREFIX: "25"                                               # Subnet Mask in CIDR notation for your node IP ranges
 DNS_SERVERS: "[8.8.8.8,8.8.4.4]"                              # The dns nameservers for the machines network-config.
 BRIDGE: "vmbr1"                                               # The network bridge device for Proxmox VE VMs
+VLAN: "2"                                                     # (optional) The VLAN to put the bridge device into
```
mcbenjemaa, line 94: There is only one remark here, 
you make the VLAN as non optional.

mcbenjemaa, line 94:
```suggestion
VLAN: "2"                                                     # The VLAN to put the bridge device into
```

mcbenjemaa: if you want, you can also set the default value in the template. so it's considered optional.

Luzifer: I think setting it mandatory would break the previous flow as existing configs will complain. Need to have a look which default works here. From `*uint16` it should be `null`… 🤔

Luzifer: Updated the templates to have a `null` default value.

mcbenjemaa: the null will not work for most cases, 
and you don't need to add VLAN to all cluster templates.

Luzifer: Why will the `null` not work? The default of an unset pointer is `null` and the type is a `*uint16`… 🤔

65278 (changes requested):
The "VLAN" field doesn't actually select or add a vlan, it only adds a tag. It's basically broken, and we should fix documentation on this (or remove). This is also why it's not in the templates.

Edit: I'll word it more clearly, this tells the bridge in proxmox (on the hypervisor) to tag with a vlan id, which only works in certain hypervisor configurations. In general you're expected to create a bridge per vlan id you want.

Luzifer:
> It's basically broken, and we should fix documentation on this (or remove). This is also why it's not in the templates.

So if it is basically broken, we can close this PR and the related issue?

65278:
It may or may not work depending on how you set up your proxmox servers, it's like tagging a port in a switch, rather than creating a vlan in the deployed machine.
I'm okay with the PR in general, but please clarify the docs on this in the CRD if you push it into the templates.

Issue #338, opened 2024-11-27
Support image lookup by tags


pborn-ionos:
Reference: https://github.com/ionos-cloud/cluster-api-provider-ionoscloud/blob/main/docs/custom-image.md#enabling-image-lookup

CAPIC already supports referencing templates by label, instead of requiring the user to supply the UUID of the image. Since Proxmox also supports tags on VMs, we should also support referencing the template by tags, rather than requiring supplying the VMID of the template VM.

wikkyk:
fixed by #343

PR #343, opened 2024-12-02
feat: allow vmtemplate selection based on tags

pborn-ionos:
* `(sourceNode + templateID)` and `templateSelector` are supposed to be mutually exclusive
* automatically detects both `sourceNode` + `templateID`
* errors out if anything but one (1) VM template with desired flags was found

*Issue #, if available:*
#338 

*Description of changes:*
Allows selecting the vm template via
```yaml
spec:
  template:
    spec:
      templateSelector:
        matchTags:
        - capi
        - capmox
        - template
        - v1.28.3
```
on `ProxmoxMachineTemplate` objects.

*Testing performed:*

- successfully provisioned machines using the existing sourceNode+templateID and the new templateSelector options
- successfully errored out, when multiple, or zero, templates with matching tags were found


api/v1alpha1/proxmoxmachine_types.go
```diff
@@ -145,6 +145,7 @@ const (
 )
 
 // VirtualMachineCloneSpec is information used to clone a virtual machine.
+// TODO +kubebuilder:validation:XValidation:rule="(self.sourceNode != ” || self.templateID != 0) && (self.templateSelector != null)",message="must define either SourceNode and TemplateID, OR TemplateSelector"
```
65278, line 148: Make sourceNode a pointer and check for it to be null.
I used that trick here: https://github.com/ionos-cloud/cluster-api-provider-proxmox/blob/main/api/v1alpha1/proxmoxmachine_types.go#L363

api/v1alpha1/proxmoxmachine_types.go
```diff
@@ -526,9 +538,20 @@ func (r *ProxmoxMachine) GetTemplateID() int32 {
 	return -1
 }
 
+// GetTemplateSelectorTags get the tags, the desired vm template should have.
+func (r *ProxmoxMachine) GetTemplateSelectorTags() []string {
+	if r.Spec.TemplateSelector != nil && r.Spec.TemplateSelector.MatchTags != nil {
+		return r.Spec.TemplateSelector.MatchTags
+	}
+	return nil
```
mcbenjemaa, line 546:
```suggestion
	return ptr.Deref(r.Spec.TemplateSelector, TemplateSelector{MatchTags: []string{}})
```

internal/service/vmservice/vm.go
```diff
@@ -371,6 +371,17 @@ func createVM(ctx context.Context, scope *scope.MachineScope) (proxmox.VMCloneRe
...
 
 	templateID := scope.ProxmoxMachine.GetTemplateID()
+	if templateID == -1 {
+		var err error
+
+		templateSelectorTags := scope.ProxmoxMachine.GetTemplateSelectorTags()
+		options.Node, templateID, err = scope.InfraCluster.ProxmoxClient.FindVMTemplateByTags(ctx, templateSelectorTags)
+
+		if err != nil {
+			scope.SetFailureMessage(err)
```
mcbenjemaa, line 381: You don't need this Failure unless you know 100% that the errors is caused by template not found!

mcbenjemaa: and please write better message with an error reason.

api/v1alpha1/proxmoxmachine_types.go
```diff
@@ -162,12 +162,16 @@ type VirtualMachineCloneSpec struct {
...
 	// +kubebuilder:validation:MinLength=1
+	// +optional
 	SourceNode string `json:"sourceNode"`
 
 	// TemplateID the vm_template vmid used for cloning a new VM.
 	// +optional
 	TemplateID *int32 `json:"templateID,omitempty"`
 
+	// +optional
+	TemplateSelector *TemplateSelector `json:"templateSelector,omitempty"`
```
mcbenjemaa, line 173: we need some API tests for this new type.

api/v1alpha1/proxmoxmachine_types.go
```diff
@@ -162,12 +162,16 @@ type VirtualMachineCloneSpec struct {
 	// will be cloned onto the same node as SourceNode.
 	//
 	// +kubebuilder:validation:MinLength=1
+	// +optional
 	SourceNode string `json:"sourceNode"`
```
mcbenjemaa, line 166: sourceNode is still required.
add omitempty

mcbenjemaa: Or is that this should be required?

internal/service/vmservice/vm.go
mcbenjemaa: missing `FailureReason`

api/v1alpha1/proxmoxmachine_types.go
mcbenjemaa: you set validation min Items, 
so add test for that case.

api/v1alpha1/proxmoxmachine_types.go
mcbenjemaa: It's not  breaking change, 


```
// VirtualMachineCloneSpec is information used to clone a virtual machine.
type VirtualMachineCloneSpec struct {

	// +kubebuilder:validation:XValidation
	TemplateSource `json:",inline"`
}


type TemplateSource struct {
	// SourceNode is the initially selected proxmox node.

	// +kubebuilder:validation:MinLength=1
	SourceNode string `json:"sourceNode"`

	// TemplateID the vm_template vmid used for cloning a new VM.
	// +optional
	TemplateID *int32 `json:"templateID,omitempty"`

	// TemplateSelector is a selector to find the template VM.
	// +optional
	TemplateSelector *TemplateSelector `json:"templateSelector,omitempty"`
}

```

mcbenjemaa: it's okay, for webhook

api/v1alpha1/proxmoxmachine_types.go
```diff
@@ -202,6 +207,14 @@ type VirtualMachineCloneSpec struct {
 	Target *string `json:"target,omitempty"`
 }
 
+// TemplateSelector defines tags for looking up images.
+type TemplateSelector struct {
+	// Specifies all tags to look for, when looking up the VM template.
+	//
+	// +kubebuilder:validation:MinItems=1
```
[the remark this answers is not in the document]
65278: Okay, we need to do that in the webhook (do we need to do that? Not really, but it would've been nice to just not eat the yaml here).
Basically, this used to be possible, but now is no longer allowed.
The reasoning seems to be that algorithmic complexity is deemed O(n²) (every element is going to be compared to every element). This is clearly stupid because an actual uniqueness test is just a sort, which is O(n*log(n)), which is much better than what Kubernetes is complaining about ...

api/v1alpha1/proxmoxmachine_types.go
mcbenjemaa: I think we need to reformat of the spec  

```

type TemplateSource struct {
	// SourceNode is the initially selected proxmox node.

	// +kubebuilder:validation:MinLength=1
	SourceNode string `json:"sourceNode"`

	// TemplateID the vm_template vmid used for cloning a new VM.
	// +optional
	TemplateID *int32 `json:"templateID,omitempty"`

	// TemplateSelector is a selector to find the template VM.
	// +optional
	TemplateSelector *TemplateSelector `json:"templateSelector,omitempty"`
}

```

so we can use it in #371

api/v1alpha1/proxmoxmachine_types_test.go
```diff
@@ -56,18 +59,24 @@ var _ = Describe("ProxmoxMachine Test", func() {
...
+			dm.Spec.TemplateSelector = &TemplateSelector{MatchTags: []string{"test"}}
+			Expect(k8sClient.Create(context.Background(), dm)).Should(MatchError(ContainSubstring("mutually exclusive")))
+		})
 
-			Expect(k8sClient.Create(context.Background(), dm)).Should(MatchError(ContainSubstring("Must set full=true when specifying format")))
+		It("Should not allow specifying TemplateSelector with empty MatchTags", func() {
+			dm := defaultMachine()
+			dm.Spec.TemplateSelector = &TemplateSelector{MatchTags: []string{}}
+
+			Expect(k8sClient.Create(context.Background(), dm)).Should(MatchError(ContainSubstring("should have at least 1 items")))
```
wikkyk, line 79: I think there should also be a test for both TemplateID and TemplateSelector being unset.

api/v1alpha1/proxmoxmachine_types.go
```diff
@@ -566,9 +585,20 @@ func (r *ProxmoxMachine) GetTemplateID() int32 {
...
+	return nil
+}
+
 // GetNode get the Proxmox node used to provision this machine.
 func (r *ProxmoxMachine) GetNode() string {
-	return r.Spec.SourceNode
+	if r.Spec.SourceNode != "" {
+		return r.Spec.SourceNode
+	}
+	return ""
```
wikkyk, line 601: I don't like this. The `if` as such is redundant (if the string is not empty then empty then return it otherwise return an empty string - could just return the string to the same effect) but the actual problem is that the data structure is wrong and you can't tell between an unset `SourceNode` and a `SourceNode` explicitly set to empty string.

This isn't much of a problem for clone specs but it can be a problem for machines as `ProxmoxMachineSpec` includes `TemplateSource`. `SourceNode` is mandatory for a machine, it is only optional for a template. Not only does change makes it incorrectly optional in all cases, it prevents validation.

`GetNode` returning an empty string isn't safe e.g. it will not work for deletion: https://github.com/ionos-cloud/cluster-api-provider-proxmox/blob/main/pkg/proxmox/goproxmox/api_client.go#L144

wikkyk (changes requested):

api/v1alpha1/proxmoxmachine_types.go
```diff
@@ -566,9 +585,20 @@ func (r *ProxmoxMachine) GetTemplateID() int32 {
 	return -1
 }
 
+// GetTemplateSelectorTags get the tags, the desired vm template should have.
+func (r *ProxmoxMachine) GetTemplateSelectorTags() []string {
+	if r.Spec.TemplateSelector != nil && r.Spec.TemplateSelector.MatchTags != nil {
+		return r.Spec.TemplateSelector.MatchTags
+	}
+	return nil
```
wikkyk, line 593: Similarly here the second nil check is redundant.

internal/service/vmservice/vm.go
```diff
@@ -384,6 +384,18 @@ func createVM(ctx context.Context, scope *scope.MachineScope) (proxmox.VMCloneRe
 	}
 
 	templateID := scope.ProxmoxMachine.GetTemplateID()
+	if templateID == -1 {
+		var err error
+
+		templateSelectorTags := scope.ProxmoxMachine.GetTemplateSelectorTags()
+		options.Node, templateID, err = scope.InfraCluster.ProxmoxClient.FindVMTemplateByTags(ctx, templateSelectorTags)
+
+		if err != nil {
```
mcbenjemaa, line 393: now the trick is here,

after you check `err != nil `
you need `if errors.Is(err, proxmox.ErrTemplateNotFound) {`

in this case you throw the Failure reason and message.

the normal case would be to retrun the error to the reconciler, so it can be requeed.

meaning, we only set Failure when the error is template not found,
other errors might occurs, e.g. network issue, proxmox api down, etc.

in that case, we need to return the error so the resource can try in the next reconciliation

api/v1alpha1/proxmoxmachine_types_test.go
```diff
@@ -56,18 +59,24 @@ var _ = Describe("ProxmoxMachine Test", func() {
...
 			dm := defaultMachine()
-			dm.Spec.SourceNode = ""
+			dm.Spec.Full = ptr.To(false)
 
-			Expect(k8sClient.Create(context.Background(), dm)).Should(MatchError(ContainSubstring("should be at least 1 chars long")))
+			Expect(k8sClient.Create(context.Background(), dm)).Should(MatchError(ContainSubstring("Must set full=true when specifying format")))
 		})
 
-		It("Should not allow specifying format if full clone is disabled", func() {
+		It("Should not allow specifying TemplateSelector with SourceNode and/or TemplateID", func() {
```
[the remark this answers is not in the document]
wikkyk: Isn't it the whole point that you define only one of (SourceNode, TemplateID)?

api/v1alpha1/proxmoxmachine_types.go
wikkyk: The underlying issue remains: `SourceNode` is unconditionally optional now, which is incorrect.

wikkyk: Ok, I see the current version leaves the `+kubebuilder:validation:MinLength=1` rule intact so at least you can't *set* it to empty.

pkg/proxmox/goproxmox/api_client.go
```diff
@@ -141,6 +142,51 @@ func (c *APIClient) FindVMResource(ctx context.Context, vmID uint64) (*proxmox.C
 	return nil, fmt.Errorf("unable to find VM with ID %d on any of the nodes", vmID)
 }
 
+// FindVMTemplateByTags tries to find a VMID by its tags across the whole cluster.
+func (c *APIClient) FindVMTemplateByTags(ctx context.Context, templateTags []string) (string, int32, error) {
```
wikkyk, line 146: Hm, `templateTags` could be `nil` here:`GetTemplateSelectorTags` can return `nil` but its result is passed to `FindVMTemplateByTags` https://github.com/ionos-cloud/cluster-api-provider-proxmox/pull/343/files#diff-1a30de7fe02a7f3127edda65e72bcd4b7a77ac737126aed0fe007d3cfba9d224R389-R390

Can you add a test that calls `FindVMTemplateByTags(..., nil)`, please?

api/v1alpha1/proxmoxmachine_types_test.go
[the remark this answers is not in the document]
wikkyk: Ok, nevermind, I misunderstood it. I think the word 'together' makes it clearer:
```suggestion
		It("Should not allow specifying TemplateSelector together with SourceNode and/or TemplateID", func() {
```

wikkyk (approved):

mcbenjemaa (commented):
Tested:

The template selector works
but,

I tested the case of a multiple template have same tags: 
`capi;capmox`

and other templates have more tags in addition.

the selector selects the VM template with `capi;capmox` as expected and ignores the others.

but Why we need to match similar tags?
I mean if a VM has more tags than that it's okay, and the selector can select it.
but if the number of VMs selected is confusing we can do something about it,.

Is that what we need?
the `matchTags` should exactly match the tags in the VM template,
if a VM has the `matchTags` and some additional tags will not be selected?

mcbenjemaa:
a typical use case:
I have a template with 

`capi;kube_v1.29.3;ubuntu `

And the selector is:

```
  templateSelector:
    matchTags:
      - kube_v1.29.3
      - ubuntu
```

the result is:


```
  Failure Message:         found 0 VM templates with tags "kube_v1.29.3;ubuntu"
  Failure Reason:          VMTemplateNotFound

```

wikkyk:
Hm indeed that's not how I'd expect selection by tag to work, either. I shouldn't need to pass the exact tags, just enough tags to get a result.

mcbenjemaa:
> Hm indeed that's not how I'd expect selection by tag to work, either. I shouldn't need to pass the exact tags, just enough tags to get a result.

what about multiple VMs has the same matchTags
what would you suggest in this case?

wikkyk:
You need to pass enough tags to get a unique result. I agree with you though that you shouldn't need to pass _all_ the tags a template has.

wikkyk:
We had a chat about this and decided to keep this behaviour and document it.

mcbenjemaa (dismissed):
Fix the lint and you are good to go.

mcbenjemaa (approved):

PR #352, opened 2024-12-05
go 1.23

wikkyk:
Go 1.23 plus the minimum necessary deps.


wikkyk:
We want capi 1.9 first.

wikkyk:
Ready for review.

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/ionos-cloud/cluster-api-provider-proxmox
 
-go 1.22.9
+go 1.23
```
mcbenjemaa, line 3: it's better to pin the go tool version. 

```
go 1.23.0
```

wikkyk: I'll let that happen through one of the dependency update PRs. In this PR I just wanted to make the minimum set of changes for go 1.23.

wikkyk: They'll just go straight for 1.23.7, anyway.

mcbenjemaa: Yeah, but will provoke some diff when someone tries to update some deps.
Cause everyone has his own go version.

wikkyk: I'lll set it it to 1.23.7 because that's what the existing [go 1.23](https://github.com/ionos-cloud/cluster-api-provider-proxmox/labels/go%2F1.23) PRs will bump it to anyway.

wikkyk:
I don't think we need to waste an e2e run on a trivial change (1.23 → 1.23.7).

mcbenjemaa (approved):

PR #360, opened 2024-12-10
Add kubernetes-version/injectKubernetesVersion to Metadata

rybnico:
*Description of changes:*

It would be very helpful to have access to the Kubernetes version specified in a machine's `spec.version` in the cloud-init/ignition metadata. With this information, it is possible to install Kubernetes version specific resources in cloud-init/ignition.

I used [this PR](https://github.com/ionos-cloud/cluster-api-provider-proxmox/pull/347) as a template, but I'm not sure if the way to control the injection is particularly elegant.
I would suggest injecting the field without a flag in the default or always passing the value of `MetadataSettings` completely to `cloudinit.NewMetadata` so that you don't have to add another argument for each flag.
I would be very happy to receive feedback from the maintainers on this suggestion and would then also carry out the desired implementation.

*Testing performed:*

We have tested this with the following `preKubeadmCommands`:

```yaml
      - MINOR_KUBERNETES_VERSION=$(echo {{ '{{ ds.meta_data.kubernetes_version }}' }} | cut -d. -f1-2 )
      - curl -fsSL https://pkgs.k8s.io/core:/stable:/${MINOR_KUBERNETES_VERSION}/deb/Release.key | gpg --dearmor -o /etc/apt/keyrings/kubernetes-apt-keyring.gpg
      - echo "deb [signed-by=/etc/apt/keyrings/kubernetes-apt-keyring.gpg] https://pkgs.k8s.io/core:/stable:/${MINOR_KUBERNETES_VERSION}/deb/ /" > /etc/apt/sources.list.d/kubernetes.list
      - apt-get update -y
      - TRIMMED_KUBERNETES_VERSION=$(echo {{ '{{ ds.meta_data.kubernetes_version }}' }} | sed 's/\./\\\\./g' | sed 's/^v//')
      - RESOLVED_KUBERNETES_VERSION=$(apt-cache madison kubelet | awk -v VERSION=${TRIMMED_KUBERNETES_VERSION} '$3~ VERSION { print $3 }' | head -n1)
      - apt-get install -y kubelet=${RESOLVED_KUBERNETES_VERSION} kubeadm=${RESOLVED_KUBERNETES_VERSION} kubectl=${RESOLVED_KUBERNETES_VERSION}
      - apt-mark hold kubelet kubeadm kubectl
```

### DIFF
--- a/internal/service/vmservice/bootstrap.go
+++ b/internal/service/vmservice/bootstrap.go
@@ -119,7 +119,7 @@ func injectIgnition(ctx context.Context, machineScope *scope.MachineScope, boots
 		Network:       nicData,
 	}
 
-	injector := ignitionISOInjector(machineScope.VirtualMachine, metadata, enricher)
+	injector := getIgnitionISOInjector(machineScope.VirtualMachine, metadata, enricher)
 	if err := injector.Inject(ctx, inject.IgnitionFormat); err != nil {
 		conditions.MarkFalse(machineScope.ProxmoxMachine, infrav1alpha1.VMProvisionedCondition, infrav1alpha1.VMProvisionFailedReason, clusterv1.ConditionSeverityWarning, err.Error())
 		return errors.Wrap(err, "ignition iso inject failed")
@@ -140,7 +140,7 @@ func defaultISOInjector(vm *proxmox.VirtualMachine, bootStrapData []byte, metada
 	}
 }
 
-func ignitionISOInjector(vm *proxmox.VirtualMachine, metadata cloudinit.Renderer, enricher *ignition.Enricher) isoInjector {
+func defaultIgnitionISOInjector(vm *proxmox.VirtualMachine, metadata cloudinit.Renderer, enricher *ignition.Enricher) isoInjector {
 	return &inject.ISOInjector{
 		VirtualMachine:   vm,
 		IgnitionEnricher: enricher,
@@ -148,7 +148,10 @@ func ignitionISOInjector(vm *proxmox.VirtualMachine, metadata cloudinit.Renderer
 	}
 }
 
-var getISOInjector = defaultISOInjector
+var (
+	getISOInjector         = defaultISOInjector
+	getIgnitionISOInjector = defaultIgnitionISOInjector
+)
 
 // getBootstrapData obtains a machine's bootstrap data from the relevant K8s secret and returns the data.
 // TODO: Add format return if ignition will be supported.
--- a/internal/service/vmservice/bootstrap.go
+++ b/internal/service/vmservice/bootstrap.go
@@ -69,13 +69,18 @@ func reconcileBootstrapData(ctx context.Context, machineScope *scope.MachineScop
 		return false, err
 	}
 
+	kubernetesVersion := ""
+	if machineScope.Machine.Spec.Version != nil {
+		kubernetesVersion = *machineScope.Machine.Spec.Version
+	}
+
 	machineScope.Logger.V(4).Info("reconciling BootstrapData.", "format", format)
 
 	// Inject userdata based on the format
 	if ptr.Deref(format, "") == ignition.FormatIgnition {
-		err = injectIgnition(ctx, machineScope, bootstrapData, biosUUID, nicData)
+		err = injectIgnition(ctx, machineScope, bootstrapData, biosUUID, nicData, kubernetesVersion)
 	} else if ptr.Deref(format, "") == cloudinit.FormatCloudConfig {
-		err = injectCloudInit(ctx, machineScope, bootstrapData, biosUUID, nicData)
+		err = injectCloudInit(ctx, machineScope, bootstrapData, biosUUID, nicData, kubernetesVersion)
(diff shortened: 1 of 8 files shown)


pkg/cloudinit/metadata.go
```diff
@@ -32,12 +35,14 @@ type Metadata struct {
 }
 
 // NewMetadata returns a new Metadata object.
-func NewMetadata(instanceID, hostname string, injectProviderID bool) *Metadata {
+func NewMetadata(instanceID, hostname string, kubernetesVersion string, injectProviderID bool, injectKubernetesVersion bool) *Metadata {
```
wikkyk, line 38: Could we do without both `kubernetesVersion` and `injectKubernetesVersion`? You could set `BaseCloudInitData.KubernetesVersionInjection` to true or false depending on whether `kubernetesVersion` is unset (say, empty string, or nil pointer).

If not, I would prefer the two args be kept together, like so
`func NewMetadata(instanceID, hostname string, injectProviderID bool, injectKubernetesVersion bool, kubernetesVersion string) *Metadata {`

rybnico: Thanks for your feedback!

I have removed the `injectKubernetesVersion` field. The `kubernetes-version` will now be injected into the metadata if `spec.version` is set on the machine.

I had to add a new argument to `NewMetadata` because I don't have access to the `MachineScope` there and don't know of any other way to get the Kubernetes version here.

What do you think of the new, much simpler implementation? I removed the draft flag because I think it is ready now. I have successfully tested the implementation on our infrastructure.

internal/service/vmservice/bootstrap.go
```diff
@@ -69,13 +69,18 @@ func reconcileBootstrapData(ctx context.Context, machineScope *scope.MachineScop
 		return false, err
 	}
 
+	kubernetesVersion := ""
+	if machineScope.Machine.Spec.Version != nil {
+		kubernetesVersion = *machineScope.Machine.Spec.Version
+	}
+
```
wikkyk, line 76: You don't seem to do anything with this value.

internal/service/vmservice/bootstrap.go
```diff
@@ -69,13 +69,18 @@ func reconcileBootstrapData(ctx context.Context, machineScope *scope.MachineScop
...
 	machineScope.Logger.V(4).Info("reconciling BootstrapData.", "format", format)
 
 	// Inject userdata based on the format
 	if ptr.Deref(format, "") == ignition.FormatIgnition {
-		err = injectIgnition(ctx, machineScope, bootstrapData, biosUUID, nicData)
+		err = injectIgnition(ctx, machineScope, bootstrapData, biosUUID, nicData, kubernetesVersion)
```
wikkyk, line 81: This line has no tests.

internal/service/vmservice/bootstrap.go
```diff
@@ -101,9 +106,9 @@ func injectCloudInit(ctx context.Context, machineScope *scope.MachineScope, boot
 	return nil
 }
 
-func injectIgnition(ctx context.Context, machineScope *scope.MachineScope, bootstrapData []byte, biosUUID string, nicData []cloudinit.NetworkConfigData) error {
+func injectIgnition(ctx context.Context, machineScope *scope.MachineScope, bootstrapData []byte, biosUUID string, nicData []cloudinit.NetworkConfigData, kubernetesVersion string) error {
 	// create metadata renderer
-	metadata := cloudinit.NewMetadata(biosUUID, machineScope.Name(), ptr.Deref(machineScope.ProxmoxMachine.Spec.MetadataSettings, infrav1alpha1.MetadataSettings{ProviderIDInjection: false}).ProviderIDInjection)
+	metadata := cloudinit.NewMetadata(biosUUID, machineScope.Name(), kubernetesVersion, ptr.Deref(machineScope.ProxmoxMachine.Spec.MetadataSettings, infrav1alpha1.MetadataSettings{ProviderIDInjection: false}).ProviderIDInjection)
```
wikkyk, line 111: This line has no tests.

wikkyk (changes requested):
I like it now. Please address the comment. I marked the two lines that sonarqube marked untested - if you add the tests, it should bring coverage to 100%.

internal/service/vmservice/bootstrap.go
rybnico: I have added tests that cover this line. It took some time and I had to make some changes.
The whole ignition bootstrapping had no real tests, I think this should have been done before.
Anyway, now we are at 88.9% coverage of the new code. As I don't have access to the SonarQube reports, I'm not sure which lines are still not covered.

internal/service/vmservice/bootstrap.go
rybnico: See my comment above

internal/service/vmservice/bootstrap.go
wikkyk: Indeed it should have been done before :)

wikkyk (approved):

PR #366, opened 2024-12-13
chore: add comment to `e2e-image` in Makefile

avivkeller:
CC @wikkyk


wikkyk:
The change requires approval now - I think that's the intention.

PR #384, opened 2025-01-21
workflows: move testing external PRs to own workflow

wikkyk:
This adds a workflow that runs tests on PRs from forks. The Test workflow instead writes a comment that reminds o/ approval and highlights potentially suspicious files.

Ref https://github.com/ionos-cloud/cluster-api-provider-proxmox/security/advisories/GHSA-qhwh-qr65-jwm3

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -3,56 +3,50 @@ on:
   push:
     branches: ["main"]
   pull_request_target:
-    types: ["opened", "synchronize", "reopened"]
     branches: ["main"]
 
 jobs:
   go_test:
+    if: ${{ github.event_name == 'push' }}
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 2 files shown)


wikkyk:
I also removed a fem IMHO superfluous lines from the Test workflow.

.github/workflows/test.yml
```diff
@@ -3,56 +3,50 @@ on:
   push:
     branches: ["main"]
   pull_request_target:
```
mcbenjemaa, line 5: I think this should be changed to `pull_request`

wikkyk: No, because then we'd run the action from the pull request not from main.

mcbenjemaa: so both actions will use pull_request_target

mcbenjemaa (approved):

Issue #425, opened 2025-02-20
Allow IPPoolRef in ProxmoxCluster


Banh-Canh:
Hello !

I'd like to be able to create one Global IPAM Pool using the required InClusterIPAM
And reference it so any of my cluster can pick IPAdresses in it.

As of now, it seems that for each ProxmoxCluster you create you have to specify the addresses ranges (using ipv4Config) which then create a dedicated IPPool.

Basically, I want multiple ProxmoxClusters to pick ip from the same pool 

Thanks for reading :)

mcbenjemaa:
thanks for reporting this feature request.

related issues #427 #426

wikkyk:
I will close this issue now as the feature is implemented and released in 0.7.0. Please feel free to reopen if necessary.

PR #429, opened 2025-02-24
Add API spec for machine ip pool ref

### LINKED ISSUE #427: Add IPPoolRef to ProxmoxMachine to override the IP range.
[A clear and concise description of what you want to happen.]

Currently, the default network spec is defined in the ProxmoxCluster spec.
Since we started provisioning clusters in various zones, we need a way to change the IP range for specific machines.

 We started #423 to place specific machines into some set of Proxmox Nodes.
 We need something similar to override the IP range for the default network while supporting backward compatibility.

**Proposal:**

Add new struct in proxmoxmachines_types.go:

```
// IPPoolConfig defines the IPAM pool ref.
type IPPoolConfig struct {
	// IPv4PoolRef is a reference to an IPAM Pool resource, which exposes IPv4 addresses.
	// The network device will use an available IP address from the referenced pool.
	// This can be combined with `IPv6PoolRef` in order to enable dual stack.
	// +optional
	// +kubebuilder:validation:XValidation:rule="self.apiGroup == 'ipam.cluster.x-k8s.io'",message="ipv4PoolRef allows only IPAM apiGroup ipam.cluster.x-k8s.io"
	// +kubebuilder:validation:XValidation:rule="self.kind == 'InClusterIPPool' || self.kind == 'GlobalInClusterIPPool'",message="ipv4PoolRef allows either InClusterIPPool or GlobalInClusterIPPool"
	IPv4PoolRef *corev1.TypedLocalObjectReference `json:"ipv4PoolRef,omitempty"`

	// IPv6PoolRef is a reference to an IPAM pool resource, which exposes IPv6 addresses.
	// The network device will use an available IP address from the referenced pool.
	// this can be combined with `IPv4PoolRef` in order to enable dual stack.
	// +optional
	// +kubebuilder:validation:XValidation:rule="self.apiGroup == 'ipam.cluster.x-k8s.io'",message="ipv6PoolRef allows only IPAM apiGroup ipam.cluster.x-k8s.io"
	// +kubebuilder:validation:XValidation:rule="self.kind == 'InClusterIPPool' || self.kind == 'GlobalInClusterIPPool'",message="ipv6PoolRef allows either InClusterIPPool or GlobalInClusterIPPool"
	IPv6PoolRef *corev1.TypedLocalObjectReference `json:"ipv6PoolRef,omitempty"`
}
```

Then adding the IPPoolConfig to the NetworkDevice:

```

// NetworkDevice defines the required details of a virtual machine network device.
type NetworkDevice struct {
       ...

	// IPPoolConfig defines config for IP Pool ref.
	// For default device 'net0' the IP pool is optional,
	// If not set, the default IPAM pool will be used.
	// For additional devices, the IP pool is required (IPV4/IPV6).
	// +optional
	IPPoolConfig `json:",inline"`
}
```

And delete the IP pool config from InterfaceConfig.

**Environment:**

- Cluster-api-provider-proxmox version:
- Kubernetes version: (use `kubectl version`):
- OS (e.g. from `/etc/os-release`):

mcbenjemaa:
This PR adds the field `IPPoolConfig ` in the ProxmoxMachine spec for the default network so that the machine can override its IP Pool.

Closes #427 
Ref: #425


wikkyk (commented):
I think it's ok. The webhook could use more tests.

PR #430, opened 2025-02-27
Enable logging for e2e tests

mcbenjemaa:
This PR enable logging for the e2e tests so we can check what's happening.


wikkyk (approved):
I think we'll want to keep this even after debugging e2e, anyway.

PR #432, opened 2025-03-02
Add CalcOnlyRunning option to ProxMox scheduler

deepdivenow:
*Issue #, if available:*

*Description of changes:*
Add CalcOnlyRunning option to ProxMox scheduler.
Add Scheduler feature ignore VMs in Not Running State for test environments.
Refactor goproxmox options for support new local options

*Testing performed:*
Testing on local env with proxmox 8.3.3

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -24,6 +24,7 @@ import (
 	"fmt"
 	"net/http"
 	"os"
+	"strings"
 
 	"github.com/go-logr/logr"
 	"github.com/luthermonson/go-proxmox"
@@ -69,7 +70,8 @@ var (
 	// ProxmoxTokenID env variable that defines the Proxmox token id.
 	ProxmoxTokenID string
 	// ProxmoxSecret env variable that defines the Proxmox secret for the given token id.
-	ProxmoxSecret string
+	ProxmoxSecret             string
+	ProxmoxReserveOnlyRunning string
 
 	proxmoxInsecure     bool
 	proxmoxRootCertFile string
@@ -208,10 +210,12 @@ func setupProxmoxClient(ctx context.Context, logger logr.Logger) (capmox.Client,
 	}
 
 	httpClient := &http.Client{Transport: tr}
-	return goproxmox.NewAPIClient(ctx, logger, ProxmoxURL,
-		proxmox.WithHTTPClient(httpClient),
-		proxmox.WithAPIToken(ProxmoxTokenID, ProxmoxSecret),
-	)
+	opts := []any{proxmox.WithHTTPClient(httpClient), proxmox.WithAPIToken(ProxmoxTokenID, ProxmoxSecret)}
+	if strings.ToLower(ProxmoxReserveOnlyRunning) == "true" {
+		opts = append(opts, goproxmox.WithCalcOnlyRunning())
+	}
+
+	return goproxmox.NewAPIClient(ctx, logger, ProxmoxURL, opts...)
 }
 
 func initFlagsAndEnv(fs *pflag.FlagSet) {
@@ -220,6 +224,7 @@ func initFlagsAndEnv(fs *pflag.FlagSet) {
 	ProxmoxURL = env.GetString("PROXMOX_URL", "")
 	ProxmoxTokenID = env.GetString("PROXMOX_TOKEN", "")
 	ProxmoxSecret = env.GetString("PROXMOX_SECRET", "")
+	ProxmoxReserveOnlyRunning = env.GetString("PROXMOX_RESERVE_ONLY_RUNNING", "")
 
 	fs.BoolVar(&proxmoxInsecure, "proxmox-insecure",
 		env.GetString("PROXMOX_INSECURE", "true") == "true",
--- a/pkg/proxmox/goproxmox/api_client.go
+++ b/pkg/proxmox/goproxmox/api_client.go
@@ -39,18 +39,40 @@ var ErrVMIDFree = errors.New("VMID is free")
 // APIClient Proxmox API client object.
 type APIClient struct {
 	*proxmox.Client
-	logger logr.Logger
+	logger             logr.Logger
+	reserveOnlyRunning bool
+}
+
+type Option func(*APIClient)
+
+func WithCalcOnlyRunning() Option {
+	return func(c *APIClient) {
+		c.reserveOnlyRunning = true
+	}
 }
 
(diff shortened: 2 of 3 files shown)


cmd/main.go
```diff
@@ -208,10 +210,12 @@ func setupProxmoxClient(ctx context.Context, logger logr.Logger) (capmox.Client,
 	}
 
 	httpClient := &http.Client{Transport: tr}
-	return goproxmox.NewAPIClient(ctx, logger, ProxmoxURL,
-		proxmox.WithHTTPClient(httpClient),
-		proxmox.WithAPIToken(ProxmoxTokenID, ProxmoxSecret),
-	)
+	opts := []any{proxmox.WithHTTPClient(httpClient), proxmox.WithAPIToken(ProxmoxTokenID, ProxmoxSecret)}
+	if strings.ToLower(ProxmoxReserveOnlyRunning) == "true" {
+		opts = append(opts, goproxmox.WithCalcOnlyRunning())
```
mcbenjemaa, line 215: Wouldn't be better to set it as field in the cluster/machine spec?

wikkyk:
What's the context for this change?

cmd/main.go
```diff
@@ -69,7 +70,8 @@ var (
 	// ProxmoxTokenID env variable that defines the Proxmox token id.
 	ProxmoxTokenID string
 	// ProxmoxSecret env variable that defines the Proxmox secret for the given token id.
-	ProxmoxSecret string
+	ProxmoxSecret             string
+	ProxmoxReserveOnlyRunning string
```
wikkyk, line 74: The name of this variable isn't particularly descriptive and it's missing a comment.
The type should be `bool`.

cmd/main.go
```diff
@@ -208,10 +210,12 @@ func setupProxmoxClient(ctx context.Context, logger logr.Logger) (capmox.Client,
 	}
 
 	httpClient := &http.Client{Transport: tr}
-	return goproxmox.NewAPIClient(ctx, logger, ProxmoxURL,
-		proxmox.WithHTTPClient(httpClient),
-		proxmox.WithAPIToken(ProxmoxTokenID, ProxmoxSecret),
-	)
+	opts := []any{proxmox.WithHTTPClient(httpClient), proxmox.WithAPIToken(ProxmoxTokenID, ProxmoxSecret)}
+	if strings.ToLower(ProxmoxReserveOnlyRunning) == "true" {
```
wikkyk, line 214: Use a `bool`.

cmd/main.go
```diff
@@ -220,6 +224,7 @@ func initFlagsAndEnv(fs *pflag.FlagSet) {
 	ProxmoxURL = env.GetString("PROXMOX_URL", "")
 	ProxmoxTokenID = env.GetString("PROXMOX_TOKEN", "")
 	ProxmoxSecret = env.GetString("PROXMOX_SECRET", "")
+	ProxmoxReserveOnlyRunning = env.GetString("PROXMOX_RESERVE_ONLY_RUNNING", "")
```
wikkyk, line 227: A new environment variable needs user documentation.

pkg/proxmox/goproxmox/api_client.go
```diff
@@ -39,29 +39,55 @@ var ErrVMIDFree = errors.New("VMID is free")
...
+func WithCalcOnlyRunning() Option {
+	return func(c *APIClient) {
+		c.reserveOnlyRunning = true
+	}
 }
 
 // NewAPIClient initializes a Proxmox API client. If the client is misconfigured, an error is returned.
-func NewAPIClient(ctx context.Context, logger logr.Logger, baseURL string, options ...proxmox.Option) (*APIClient, error) {
+// options can be goproxmox.Option or proxmox.Option
+func NewAPIClient(ctx context.Context, logger logr.Logger, baseURL string, options ...any) (*APIClient, error) {
```
wikkyk, line 56: Definitely not `...any`.

pkg/proxmox/goproxmox/api_client.go
```diff
@@ -39,29 +39,55 @@ var ErrVMIDFree = errors.New("VMID is free")
...
+	for _, option := range options {
+		switch o := option.(type) {
+		case proxmox.Option:
+			pOpts = append(pOpts, o)
+		case Option:
+			apiOpts = append(apiOpts, o)
+		default:
+			return nil, fmt.Errorf("invalid option %T", option)
+		}
+	}
```
wikkyk, line 68: This should've been a type.

pkg/proxmox/goproxmox/api_client.go
```diff
@@ -39,29 +39,55 @@ var ErrVMIDFree = errors.New("VMID is free")
...
-	return &APIClient{
+	ac := &APIClient{
 		Client: upstreamClient,
 		logger: logger,
-	}, nil
+	}
+	for _, o := range apiOpts {
+		o(ac)
+	}
+	return ac, nil
```
wikkyk, line 90: Ditto.

cmd/main.go
wikkyk: I would tend to agree.

wikkyk (changes requested):
I don't like this change. I _think_ I get the use case but you need to elaborate on the problem being solved here.
Besides, the naming is inconsistent and it's a user interface change so it needs user documentation.

pkg/proxmox/goproxmox/api_client.go
deepdivenow: Can you get example
How to not use any, if we have type from upstream library "proxmox.Option", and we need local type in this library?

cmd/main.go
deepdivenow: > Wouldn't be better to set it as field in the cluster/machine spec?
Maybe on Cluster scope, on mashine scope need special config structure for each physical node in physical proxmox Cluster.  
proxmoxmachines.infrastructure.cluster.x-k8s.io is virtualMashine and very strange set this option here (not For Physical mashine|node)

deepdivenow:
> What's the context for this change?

For example my test environment have a lot VMs, but most of the time these machines are not working.
Now scheduler calc RAM of this mashines as USED, but physical node have a lot of free memory for run new VMs.
изображение

pkg/proxmox/goproxmox/api_client.go
wikkyk: You can either create a new type that's composed of the two types or extend proxmox.Option since it's our own type.

PR #441, opened 2025-03-10
.github/workflows/{e2e,test}: check out sha instead of ref

wikkyk:
`ref` can be raced between approval and checkout by pushing a commit between approval and checkout. `sha` always refers to the triggering commit, which is safe.

*Issue #, if available:*
https://github.com/ionos-cloud/cluster-api-provider-proxmox/security/advisories/GHSA-qhwh-qr65-jwm3


wikkyk:
Adding e2e/none as running the e2e tests won't do anything here.

mcbenjemaa (approved):

PR #483, opened 2025-05-09
Bump golangci-lint

mcbenjemaa:


.github/workflows/lint.yml
```diff
@@ -17,9 +17,10 @@ jobs:
         with:
           go-version-file: go.mod
       - name: golangci-lint
-        uses: golangci/golangci-lint-action@v7
+        uses: golangci/golangci-lint-action@v8
         with:
-          version: v2.0.2
+          version: v2.1.6
+          args: '--timeout 10m'
```
wikkyk, line 23: Why?

.gitignore
```diff
@@ -47,3 +47,5 @@ _artifacts
 test/e2e/config/*-envsubst.yaml
 
 .tool-versions
+e2e-env*.yaml
+
```
wikkyk, line 51: intended?

mcbenjemaa: yes

.github/workflows/lint.yml
mcbenjemaa: the deadline was part of the config, since the new golang-lint action, we need the timeout to be in args

wikkyk (approved):

PR #467, opened 2025-03-25
Allow multiple nodes have templates (use local storage)

Atoms:
*Issue #, if available:*
fixes #451 

- Adjusted order of how scheduling happens
    - First we determine if we have usable template (fail fast if not)
    
- If we have usable template, determine how we need to schedule
    - if `Target` is provided schedule on that node if template is there
    - if `Target` and `allowedNodes` missing assume we want all nodes to use in scheduling (discover all nodes)
    - if templates are on more nodes then is in `allowedNodes` schedule only on `allowedNodes`
    - if templates missing on some `allowedNodes` still continue to schedule on nodes where template exists.

**LocalStorage**:

- If localstorage (we have no easy way to determine if template is no localstorage):
    - multiple templates can be found
    - assign options.Node same as `options.Target` as we will use template from same node.
    - templateID and SourceNode cannot be used in this scenario
- If !localstorage
    - only 1 template can be found, if multiple found fail
    - schedule vm on any allowedNode (either Target or allowedNodes or discovered nodes)
    - templateID and SourceNode only can be used in this scenario

Fixed existing tests to reflect multiple scenarios (moving out `templateID` and `SourceNode` from `setupReconcilerTest`

- fixed yamlfmt

`createVM` had became too complex (gocyclo) as i've written in code for now it should be function to revisit and get into smaller chunks.

Issue #304, opened 2024-10-18
add VIP_IP_RANGES


lknite:
**Describe the solution you'd like**
Similar to NODE_IP_RANGES, would define ip ranges which can be used via ipam for allocating a vip ip.

**Anything else you would like to add:**
The exact work solved by ipam for node ip addresses is required to setup a vip.  A human must locate an ip, assign it to the vip, and when no longer in use, somehow recover the ip so it can be used again in the future.  This work is already solved by ipam, we just need to use that solution for vip also.

A separate inclusterippool for vips would make implementation easier, but if the same ippool could be used for both node ips and vip ips that would also be a good feature, maybe do that by default if a VIP pool isn't defined, otherwise use the VIP pool.

**Note**
I put in a feature request at cluster-api-ipam to establish a standard for using ipam for vips also.
https://github.com/kubernetes-sigs/cluster-api-ipam-provider-in-cluster/issues/297

lknite:
https://github.com/kubernetes-sigs/cluster-api/issues/11313

wikkyk:
I think https://github.com/kubernetes-sigs/cluster-api-ipam-provider-in-cluster/issues/297 really answers your question. CAPMOX doesn't seem to be the right place to handle this. I would like to close this issue unless you have anything to add.

65278:
I'll also weigh in:
I plain do not understand your proposal. When you say cluster-api-vips, what do you actually mean. There is no such thing.

My reading allows two interpretations.
1. You mean kube-vip. It is true that we give kube-vip a single ip and we manage that. We could use more flexibility here, but I fail to see what this has to do with your proposal.
2. You mean the cluster internal IPs. These have nothing to do with capi and are managed by the clusters own IPAM. This is for good reason, as a failure of capmox would mean a failure of clusters managed by capmox. This should never happen.

65278:
One more thought: If you must manage cluster internal IPs via capmox, capi has a mechanism called ClusterResourceSet, which applies configmaps to a cluster. These can configure your cluster internal CNIs and kube-proxy to use the IP ranges you want. It's not a cooperative approach though. If you want cooperation, then you need to write your own IPAM controller.

pborn-ionos:
I think the OP only referred to assignment of kube-vip IPs via CAPMOX. Not cluster-internal :)

lknite:
[...]

Issue #293, opened 2024-10-02
RFC: v1alpha2 API


65278:
v1alpha1 has problems in the networking department which will require breaking changes.

Specifically, ProxmoxMachines have a split of default network device and additional network devices for no reason that I can discern. This duplicates code pathes everywhere and actually removes possibilities (due to how network interfaces are structured in the API, default doesn't have options that other interfaces have).
On top of that, the split of IPv4 and IPv6 pools is arbitrary (makes no sense), and we can simply get away with having an array of IPPools. This also allows multiple IPPools per interface, a feature which we can not support yet. On top of that, we possibly want to support network interfaces without IPPools attached (DHCP, BGP exclusively).

For now I'd change the API of ProxmoxMachines in the following way:

```yaml
apiVersion: infrastructure.cluster.x-k8s.io/v1alpha1
kind: ProxmoxMachineTemplate
metadata: 
  name: test-cluster-control-plane
  namespace: default
spec:   
  template:
    spec:
      disks:
        bootVolume:
          disk: scsi0
          sizeGb: 100
      format: qcow2
      full: true
      memoryMiB: 4096
      network:
        networkDevices:
         -  bridge: vmbr128
            model: virtio
            mtu: 9000
        - bridge: vmbr129
          dnsServers:
          - 10.4.1.1
          ipPoolRef:
           -  apiGroup: ipam.cluster.x-k8s.io
              kind: GlobalInClusterIPPool
              name: shared-int-service-v4-inclusterippool
           -  apiGroup: ipam.cluster.x-k8s.io
              kind: GlobalInClusterIPPool
              name: shared-int-service-v6-inclusterippool
          model: virtio
          name: net1
          linkMtu: 9000
      numCores: 4
      numSockets: 2
      sourceNode: proxmox1
      templateID: 100
```

I would resolve the conflicts about the first network device by appending the `InclusterIPPool` in front of the `ipPoolRef` array. There's no need for the first device to be called `net0`, but we can make the webhook add names automatically as per their position in the array.

Regarding `ProxmoxCluster`, I'd leave the API as is, because a Cluster can only have an IPv4 and/or an IPv6 pool. Potentially, we want to add DHCP there, but that's not a breaking change as far as I can see. We can discuss turning IPv4Pool and IPv6Pool into an actual pool reference that we supply with the cluster, which I think is the better choice, but this would be just breakage for breakages sake (we only get better validation on the to be created pool).

[...]

PR #531, opened 2025-08-11
Introduce API v1alpha2 

wikkyk:
*Issue #, if available:*
#293

*Description of changes:*
I forward-ported https://github.com/ionos-cloud/cluster-api-provider-proxmox/commit/1d76694132b1ddb548c03d5b94880695db205883 to main. I ported all the new features to v1alpha2.

~This is still very much WIP, it doesn't actually work, tests fail, etc.~ However, that's the bulk of the large changes done and we'll be able to work more incrementally now.

We'll need to keep rebasing this on main if necessary until it's ready to merge. 

We'll ignore SonarQube's complaints of code duplication, some code will be duplicated as only parts of the API are changing.

*Testing performed:*
It compiles.
~Some Many Go tests succeed, many some Go tests fail.~ We keep tests working now.
~Nothing else was tried yet.~
e2e/none label applied to reduce github spam.

TODOs (not exhaustive, some may be optional/obsolete):
- [x] Port to CAPI 1.11 #603 
- [ ] Port to CAPI v1beta2 #605
- [x] Port to controller-runtime 0.21 #603
- [x] Port to k8s 33 #603
- [ ] Bump kubeapilinter
- [ ] Update e2e to k8s 1.33
- [ ] Build container images with k8s 1.33
- [x] Stop using https://pkg.go.dev/inet.af/netaddr as it is deprecated
- [ ] Stop using github.com/pkg/errors as it is unmaintained
- [ ] Centralise checking if an address is IPv6 and consider adding a type for IP addresses
- [x] Make sure the imported name of github.com/ionos-cloud/cluster-api-provider-proxmox/api/v1alpha2 is consistently `infrav1`
- [ ] Audit all usage of ptr.Deref and consider whether to replace with validator rules/default values and then simply deref pointers
- [ ] #304 ?
- [ ] #467 / #532
- [ ] #601 ?
- [ ] Conversion webhook
- [x] Multiple IP addresses per pool (needs a reference counter)
- [ ] e2e pass
- [ ] test coverage%
- [x] Lint
- [ ] Fluff like update copyright years

### DIFF
--- a/api/v1alpha1/proxmoxcluster_types.go
+++ b/api/v1alpha1/proxmoxcluster_types.go
@@ -90,7 +90,6 @@ type ProxmoxClusterSpec struct {
 // ProxmoxClusterCloneSpec is the configuration pertaining to all items configurable
 // in the configuration and cloning of a proxmox VM.
 type ProxmoxClusterCloneSpec struct {
-	// +kubebuilder:validation:XValidation:rule="has(self.controlPlane)",message="Cowardly refusing to deploy cluster without control plane"
 	ProxmoxMachineSpec map[string]ProxmoxMachineSpec `json:"machineSpec"`
 
 	// SshAuthorizedKeys contains the authorized keys deployed to the PROXMOX VMs.
--- a/api/v1alpha1/proxmoxcluster_types_test.go
+++ b/api/v1alpha1/proxmoxcluster_types_test.go
@@ -88,17 +88,6 @@ func defaultCluster() *ProxmoxCluster {
 				Metric:    func() *uint32 { var a uint32 = 123; return &a }(),
 			},
 			DNSServers: []string{"1.2.3.4"},
-			CloneSpec: &ProxmoxClusterCloneSpec{
-				ProxmoxMachineSpec: map[string]ProxmoxMachineSpec{
-					"controlPlane": {
-						VirtualMachineCloneSpec: VirtualMachineCloneSpec{
-							TemplateSource: TemplateSource{
-								SourceNode: "pve1",
-							},
-						},
-					},
-				},
-			},
 		},
 	}
 }
@@ -159,15 +148,6 @@ var _ = Describe("ProxmoxCluster Test", func() {
 
 	It("Should allow creating valid clusters", func() {
 		Expect(k8sClient.Create(context.Background(), defaultCluster())).To(Succeed())
-	})
-
-	Context("CloneSpecs", func() {
-		It("Should not allow Cluster without ControlPlane nodes", func() {
-			dc := defaultCluster()
-			dc.Spec.CloneSpec.ProxmoxMachineSpec = map[string]ProxmoxMachineSpec{}
-
-			Expect(k8sClient.Create(context.Background(), dc)).Should(MatchError(ContainSubstring("control plane")))
-		})
 	})
 
 	Context("IPV6Config", func() {
--- a/api/v1alpha2/proxmoxcluster_types.go
+++ b/api/v1alpha2/proxmoxcluster_types.go
@@ -142,7 +142,6 @@ type ProxmoxClusterClassSpec struct {
 // in the configuration and cloning of a proxmox VM.
 type ProxmoxClusterCloneSpec struct {
 	// machineSpec is the map of machine specs.
-	// +kubebuilder:validation:XValidation:rule="self.exists_one(x, x.machineType == \"controlPlane\")",message="Cowardly refusing to deploy cluster without control plane"
 	// +listType=map
 	// +listMapKey=machineType
 	// +optional
--- a/api/v1alpha2/proxmoxcluster_types_test.go
+++ b/api/v1alpha2/proxmoxcluster_types_test.go
@@ -89,23 +89,6 @@ func defaultCluster() *ProxmoxCluster {
 				Metric:    ptr.To(int32(123)),
 			},
 			DNSServers: []string{"1.2.3.4"},
-			CloneSpec: &ProxmoxClusterCloneSpec{
-				ProxmoxClusterClassSpec: []ProxmoxClusterClassSpec{{
-					MachineType: "controlPlane",
-					ProxmoxMachineSpec: ProxmoxMachineSpec{
-						VirtualMachineCloneSpec: VirtualMachineCloneSpec{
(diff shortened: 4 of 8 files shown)


wikkyk:
I split the compat table in two and removed redundant information to make them a bit more readable. The old table was getting long.

wikkyk:
Should get https://github.com/kubernetes-sigs/kube-api-linter set up for this.
edit: done

wikkyk:
Needs porting to CAPI 1.11/v1beta2.

mcbenjemaa:
> Needs porting to CAPI 1.11/v1beta2.

You can finish your rebase and changes, and then we can migrate CAPI v1beta2 and conversation webhooks to a separate PR.

I can take this.

wikkyk:
I agree. Though we should minimise the amount of time that `main` is in this part-way-in state.

pkg/ignition/enrich.go
```diff
@@ -25,6 +25,7 @@ import (
 	ignition "github.com/flatcar/ignition/config/v2_3"
 	ignitionTypes "github.com/flatcar/ignition/config/v2_3/types"
 	"github.com/pkg/errors"
+	"inet.af/netaddr"
```
wikkyk: https://pkg.go.dev/inet.af/netaddr is deprecated

mcbenjemaa: let's create a follow-up issue for this.

wikkyk: yeah, I also don't like github.com/pkg/errors

wikkyk:
We should centralise checking if an address is IPv6. In fact, we should have a type for IP addresses.

wikkyk:
TODO: Make the imported name of github.com/ionos-cloud/cluster-api-provider-proxmox/api/v1alpha2 consistent.
edit: done

wikkyk:
I'm not too worried about sonarqube complaining about duplicated code too much right now; however, I did have this idea to simply import the bits of v1alpha1 into v1alpha2 that remain unchanged. Also makes it clear exactly what *has* changed.

wikkyk:
Should look at all usages of ptr.Deref and replace with validator rules/default values as necessary.

wikkyk:
TODO: capi v1beta2

wikkyk:
rebased on v0.7.5

wikkyk:
The forked capi v1beta1 utils are dragging down coverage% but that is a temporary fork so we shouldn't be too bothered by it.

internal/webhook/proxmoxmachine_webhook.go
```diff
@@ -264,3 +239,33 @@ func validateNetworkDeviceMTU(device *infrav1.NetworkDevice) error {
...
+
+	for _, networkDevice := range machine.Spec.Network.NetworkDevices {
+		defaultIPv4Count += b2i(networkDevice.DefaultIPv4)
+		defaultIPv6Count += b2i(networkDevice.DefaultIPv6)
+	}
+
+	if defaultIPv4Count == 0 {
+		machine.Spec.Network.NetworkDevices[0].DefaultIPv4 = ptr.To(true)
+	}
+	if defaultIPv6Count == 0 {
```
mcbenjemaa, line 296: What if we don't have IPv6? This will always set it to true.

internal/webhook/proxmoxmachine_webhook.go
```diff
@@ -264,3 +239,33 @@ func validateNetworkDeviceMTU(device *infrav1.NetworkDevice) error {
...
+	defaultIPv4Count := 0
+	defaultIPv6Count := 0
+
+	for _, networkDevice := range machine.Spec.Network.NetworkDevices {
+		defaultIPv4Count += b2i(networkDevice.DefaultIPv4)
+		defaultIPv6Count += b2i(networkDevice.DefaultIPv6)
+	}
+
+	if defaultIPv4Count == 0 {
+		machine.Spec.Network.NetworkDevices[0].DefaultIPv4 = ptr.To(true)
```
mcbenjemaa: Can we instead get the NetworkDevice by name rather than by index 0?

wikkyk: I don't think it matters here specifically.

mcbenjemaa: What if the first item is not the default device, another net like `net1`.
If we want the first item to be the default device, then we have to enforce it with the webhook.

65278: If you want something else, you need to set the tag. It's as simple as that.
I can write a map to index function, that's not too bad an idea.

65278: I've mapped this to infrav1.DefaultNetworkDevice, that seemed the most proper.

65278:
I've made a branch with NetName no longer being a pointer (it's a required field, since it's a map key).
https://github.com/ionos-cloud/cluster-api-provider-proxmox/tree/v1alpha2_netname_cleanup
This conflicts with migration hook development, so I'll merge it after Mohamed finishes.

65278:
I've noticed that ignition is currently broken (since it puts default IPs into the environment). I'll spend my efforts there.

internal/service/vmservice/ip.go
```diff
@@ -19,192 +19,236 @@ package vmservice
...
+	// TODO: This datastructure is less bad, but still bad
+	netPoolAddresses := make(map[string]map[corev1.TypedLocalObjectReference][]ipamv1.IPAddress)
 
-	// default device.
-	if requeue, err = handleDefaultDevice(ctx, machineScope, addresses); err != nil || requeue {
-		return true, errors.Wrap(err, "unable to handle default device")
+	if pm.Spec.Network != nil {
+		if requeue, err = handleDevices(ctx, machineScope, netPoolAddresses); err != nil || requeue {
+			if err == nil {
+				return true, errors.Wrap(err, "requeuing network reconcillation")
```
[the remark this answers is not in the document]
wikkyk: While copilot's suggested fix is wrong, it does bring up an important issue (and it's one of the reasons as for why I dislike github.com/pkg/errors) - errors.Wrap(nil, "foo") returns nil, not a new error "foo" that wraps nil. As such, error "requeuing network reconcillation" never gets returned, only nil, as copilot correctly pointed out.

PR #522, opened 2025-07-24
Introduce AdditionalVolumes in the Machine Spec

mcbenjemaa:
This PR adds:

* `additionalVolumes` field to the ProxmoxMachine spec

Closes #504

PR #561, opened 2025-10-07
Additional volumes

### LINKED ISSUE #504: Multiple disks
Ability to add extra disks to ProxmoxMachines and ProxmoxMachineTemplates beyond the `...spec.disks.bootVolume`.

Currently can only attach a single disk (for the bootVolume). Other providers allow creating additional disks for data storage. Eg vSphere has:
[VSphereMachineTemplate.spec.template.spec.dataDisks[]](https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/blob/8c01993dfad06a109a50ad36af874cd2795b41e0/config/default/crd/bases/infrastructure.cluster.x-k8s.io_vspheremachinetemplates.yaml#L847-L877) 
[VSphereMachine.spec.dataDisks[]](https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/blob/main/config/default/crd/bases/infrastructure.cluster.x-k8s.io_vspheremachines.yaml#L977-L1007)
GCP has:
 [GCPMachineTemplate.spec.template.spec.additionalDisks[]](https://github.com/kubernetes-sigs/cluster-api-provider-gcp/blob/ca64a1b9d75e5e4a1d390d2f01b4589518510581/config/crd/bases/infrastructure.cluster.x-k8s.io_gcpmachinetemplates.yaml#L77-L170) 
[GCPMachine.spec.additionalDisks[]](https://github.com/kubernetes-sigs/cluster-api-provider-gcp/blob/ca64a1b9d75e5e4a1d390d2f01b4589518510581/config/crd/bases/infrastructure.cluster.x-k8s.io_gcpmachines.yaml#L65-L156)
AWS has:
[AWSMachineTemplate.spec.template.spec.nonRootVolumes[]](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/4b2f770fd42cd92c21d172198a55ce0212806b18/config/crd/bases/infrastructure.cluster.x-k8s.io_awsmachinetemplates.yaml#L242-L286)
[AWSMachine.spec.nonRootVolumes[]](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/4b2f770fd42cd92c21d172198a55ce0212806b18/config/crd/bases/infrastructure.cluster.x-k8s.io_awsmachines.yaml#L229-L272)

etc. These are lists that take multiple disks. In this provider, there is only a single:
[ProxmoxMachine.spec.disks.bootVolume](https://github.com/ionos-cloud/cluster-api-provider-proxmox/blob/4ae6387cad84c01060fa966e419a80693d8f699c/config/crd/bases/infrastructure.cluster.x-k8s.io_proxmoxmachines.yaml#L95-L128) 
[ProxmoxMachineTemplate.spec.template.spec.disks.bootVolume](https://github.com/ionos-cloud/cluster-api-provider-proxmox/blob/4ae6387cad84c01060fa966e419a80693d8f699c/config/crd/bases/infrastructure.cluster.x-k8s.io_proxmoxmachinetemplates.yaml#L107-L140)
object. 

I suggest a non-breaking solution of keeping the ...spec.disks.bootVolume object, and create a ...spec.disks.additionalVolumes[] list. So will be able to specify, eg:
```
spec:
  disks:
    bootVolume:
      disk: boot 
      sizeGb: 100
    additionalVolumes:
      - disk: data-0 
         sizeGb: 100
      - disk: data-1 
         sizeGb: 50
```

The alternative is to scrap ...spec.disks.bootVolume since there will always be a bootVolume, move its properties to the root of ..spec. Ie just it's diskGiB, etc like GCP & vSphere. Then change ...spec.disks from an object to a list of extra disks. But while a simpler structure, would be a breaking change. 

**Environment:**
Cluster-api-provider-proxmox version: 0.7.1

holmesb:
Closes #504.

Description of changes:
Allows extra disks to be specified. Builds upon work done in @mcbenjemaa's [draft](https://github.com/ionos-cloud/cluster-api-provider-proxmox/pull/522). 

Unlike VMs, which have a unique ID for tracking, PVE provides no server-side ID for configuration items like disks. So this required implementing a pending queue ("pending guard") to avoid duplicates. Without this, spurious "Unused Disks" are created. Is off by default, but enabled in vm.go.

With this PR, we support disk flags: format, discard, iothread, and ssd in additionalVolumes.


api/v1alpha1/proxmoxmachine_types.go
```diff
@@ -140,27 +140,37 @@ type Storage struct {
...
 	// Size defines the size in gigabyte.
-	//
 	// As Proxmox does not support shrinking, the size
 	// must be bigger than the already configured size in the
 	// template.
-	//
 	// +kubebuilder:validation:Minimum=5
 	SizeGB int32 `json:"sizeGb"`
+	// Storage is an optional per-volume Proxmox storage name (e.g., "local-lvm", "nfs-data").
+	// If omitted, falls back to the machine's .spec.storage.
```
mcbenjemaa: this is not true, spec.storage is optional as well, did you mean falling back to the default storage of the VM template or spec.storage?

api/v1alpha1/proxmoxmachine_types.go
```diff
@@ -140,27 +140,37 @@ type Storage struct {
...
+	// +optional
+	Storage *string `json:"storage,omitempty"`
+	// Format is optional:
+	Format *TargetFileStorageFormat `json:"format,omitempty"`
```
mcbenjemaa, line 172: Format must be an enum with possbile values `raw, qcow2, vmdk`

api/v1alpha1/proxmoxmachine_types.go
```diff
@@ -140,27 +140,37 @@ type Storage struct {
...
+	// Discard enables TRIM/UNMAP support for this virtual disk.
+	// Safe on IDE/SATA/SCSI/VirtIO; maps to Proxmox "discard=on".
+	// If omitted or false, the flag is not set.
+	// +optional
+	Discard  *bool `json:"discard,omitempty"`
+	Iothread *bool `json:"iothread,omitempty"`
```
mcbenjemaa:
```suggestion
    // IOThread enables the option IO Thread, 
    // With IO Thread enabled, QEMU creates one I/O thread per storage controller rather than handling all I/O in the main event loop or vCPU threads.
    // The option IO Thread can only be used when using a disk with the VirtIO controller, or with the SCSI controller
	IOThread *bool `json:"ioThread,omitempty"`
```

api/v1alpha1/proxmoxmachine_types.go
```diff
@@ -140,27 +140,37 @@ type Storage struct {
...
+	SSD      *bool `json:"ssd,omitempty"`
```
mcbenjemaa:
```suggestion
    // SSDEmulation enables SSD emulation feature
    //  SSD emulation  sets a drive to be presented to the guest as a solid-state drive rather than a rotational hard disk
    // There is no requirement that the underlying storage actually be backed by SSDs; this feature can be used with physical media of any type
    // SSD emulation is not supported on VirtIO Block drives.
	SSDEmulation  *bool `json:"ssdEmulation,omitempty"`
```

api/v1alpha1/proxmoxmachine_types.go
```diff
@@ -140,27 +140,37 @@ type Storage struct {
...
 	// must be bigger than the already configured size in the
 	// template.
-	//
 	// +kubebuilder:validation:Minimum=5
 	SizeGB int32 `json:"sizeGb"`
+	// Storage is an optional per-volume Proxmox storage name (e.g., "local-lvm", "nfs-data").
+	// If omitted, falls back to the machine's .spec.storage.
+	// +optional
+	Storage *string `json:"storage,omitempty"`
+	// Format is optional:
```
mcbenjemaa, line 170: More docs for the format field.

internal/service/vmservice/vm.go
```diff
@@ -287,6 +327,120 @@ func reconcileVirtualMachineConfig(ctx context.Context, machineScope *scope.Mach
 		}
 	}
 
+	// Additional data disks
```
mcbenjemaa: Perhaps, if you can create another function for additional volumes as this raises complexity in the `reconcileVirtualMachineConfig`.

mcbenjemaa (changes requested):
thank you for raising this PR.

This is my initial review.
Please address the points i mentioned, until I find more time to look at the reconciliation function.

api/v1alpha1/proxmoxmachine_types.go
holmesb: Yes spec.storage of the ProxmoxMachine, PMT, or ProxmoxCluster. Or if that doesn't exist, the default storage of the PVE node. I've clarified this.

api/v1alpha1/proxmoxmachine_types.go
holmesb: Nice comment, but Proxmox field is [ssd](https://pve.proxmox.com/pve-docs/qm.1.html), not ssdEmulation. Do you agree CRDs and CAPMOX settings should align with PVE terminology? Otherwise we'll have to explain in documentation that "ssdEmulation" = the PVE "ssd" setting. Def "emulation" should be used in documentation tho, and I've included your code comment.

wikkyk:
Thanks for this. I've put hold on it because we're working on v1alpha2 now. I'll help you port it when it's released which should be Soon™. We are wanting to improve the storage config in v1alpha3 and this would fit in.
However, the approach with AdditionalVolumes is wrong. We've made the same mistake with AdditionalNetworks in v1alpha1. This approach proved too inflexible and very costly to change (it's why v1alpha2 is taking so long). The correct way to do this is to follow the zero one infinity rule. Just make it so that you can have multiple volumes.

holmesb:
ZOI redesign sounds a way off. I get the benefit of making disks uniform tho: simpler representation in schema and less documentation. No need for a second PR to make available features like discard, ioThread, ssd, etc to bootVolume. But trillion dollar clouds don't support this uniform approach, they separate the boot from data volumes the same as this PR. Eg:
AWS has [spec.rootVolume](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/4b2f770fd42cd92c21d172198a55ce0212806b18/config/crd/bases/infrastructure.cluster.x-k8s.io_awsmachinetemplates.yaml#L299-L339). [spec.nonRootVolumes[]](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/4b2f770fd42cd92c21d172198a55ce0212806b18/config/crd/bases/infrastructure.cluster.x-k8s.io_awsmachinetemplates.yaml#L242-L286) are separate.
GCP has [spec.rootDevice*](https://github.com/kubernetes-sigs/cluster-api-provider-gcp/blob/ca64a1b9d75e5e4a1d390d2f01b4589518510581/config/crd/bases/infrastructure.cluster.x-k8s.io_gcpmachinetemplates.yaml#L77-L170) fields. [spec.additionalDisks[]](https://github.com/kubernetes-sigs/cluster-api-provider-gcp/blob/ca64a1b9d75e5e4a1d390d2f01b4589518510581/config/crd/bases/infrastructure.cluster.x-k8s.io_gcpmachinetemplates.yaml#L77-L170) are separate.
vSphere has a [spec.diskGiB](https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/blob/8c01993dfad06a109a50ad36af874cd2795b41e0/config/default/crd/bases/infrastructure.cluster.x-k8s.io_vspheremachinetemplates.yaml#L847-L877) for the boot disk only. [dataDisks ](https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/blob/8c01993dfad06a109a50ad36af874cd2795b41e0/config/default/crd/bases/infrastructure.cluster.x-k8s.io_vspheremachinetemplates.yaml#L847-L877)are separate.

They all **do** support multiple disks (ie data disks). This is much more an MVP feature than disk uniformity, as shown by this being the most [up-ticked](https://github.com/ionos-cloud/cluster-api-provider-proxmox/issues/504) issue in this project.

As an interim, I propose we mark `spec.disks.additionalVolumes` Deprecated in godoc/CRD descriptions, with a note that it will be replaced by `spec.disks.volumes` in a future API (v1alpha3). My latest commit just now achieves. Merging with deprecation notices will unblock a widely requested feature and is non-breaking (no breaking change in v1alpha1). We'll still converge to ZOI volumes[] in the upcoming API, which will also bring the flags (discard/ioThread/ssd, etc.) uniformly to the boot disk.

When v1alpha3 starts, we can introduce `spec.disks.volumes[]` with `boot: true` on one entry. I can contribute conversion: v1alpha1 → v1alpha3 (map bootVolume to volumes[{boot:true,…}], and additionalVolumes to volumes[{boot:false,…}]).

I’ve pushed the deprecation notes into this PR so godoc and CRD OpenAPI have `additionalVolumes` as Deprecated with a pointer to the future `spec.disks.volumes`. We've been using this PR for many weeks without issue, so hopefully it's mergeable more-or-less as-is, but I’m happy to iterate on wording. I’m also happy to help on the v1alpha3 volumes[] impl to ensure a smooth transition.

Issue #583, opened 2025-11-10
Template selection by a subset of tags


holmesb:
Hi,

**Describe the solution you'd like**
Docs say: "Passed tags must be an exact 1:1 match with the tags on the template you want to use".

I would like to match subset of tags. Ie if I have a template on every node tagged:
capmox; k8s-1-33; ubuntu24-04; containerd-1-2-3; cni-3-4-5
and specify env var: `TEMPLATE_TAGS="capmox,k8s-1-33,ubuntu24-04"`
then CAPMOX should pick the single eligible template, regardless that I didn't also specify `containerd-1-2-3` nor `cni-3-4-5`.

This will allow the use of Proxmox tags purely for metadata, not for selection. Allows logic like: give me the Ubuntu 24.04 CAPMOX template for Kubernetes 1.33 and I don't care what version of containerd nor CNI it is. As long as this resolves to a single template, it shouldn't matter that I didn't specify all the tags. But still allows me to know the version of containerd, etc on the template should I need this info.

wikkyk:
Forgot to close.

PR #584, opened 2025-11-10
Template selection by a subset of tags

holmesb:
*Issue #583

*Description of changes:*
Make VM templates selectable by a subset of tags. Also made the tag search case-insensitive and de-duped both the wanted and present tag lists.

*Testing performed:*
Updated tests.


pborn-ionos:
Thank you for your contribution. We should be careful with breaking API changes.

Please have a look at our friends over at [CAPIC](https://github.com/ionos-cloud/cluster-api-provider-ionoscloud/). They've hidden this behind the flag `ResolutionPolicy`, which defaults to `Exact` (which is our current behaviour). You could then, for example, put this feature under `ResolutionPolicy: Subset`.

holmesb:
Agreed @pborn-ionos. Done. I took the opp to bring tag template selection to the cluster classes.

wikkyk:
We're focusing on v1alpha2 now so I won't merge any changes to v1alpha1. I'll put this on hold and we'll revisit it once v1alpha2 is released.
However, I think this is a nice change in principle, though I haven't reviewed the implementation yet.

wikkyk:
Obsoleted by #680. Thank you for your contribution!

PR #639, opened 2026-02-18
CAPI v1beta2 contract support and KAL alignment

mcbenjemaa:
## Summary
- Migrate to CAPI v1beta2 contract: metav1.Condition, initialization status, ProviderID as value type
- Separate kube-api-linter into dedicated `.golangci-kal.yml` config with `make lint-api` target (vSphere pattern)
- Fix all KAL lint issues: MaxLength/MaxItems markers, forbidden +default removal with Go-level defaulting
- Align CRD labels with v1beta2 contract (`cluster.x-k8s.io/v1beta2: v1alpha2`)

## Test plan
- [x] `make test` — all tests pass
- [x] `make lint-api` — 0 issues
- [x] Fuzzy conversion tests pass (spoke↔hub roundtrip)

PR #648, opened 2026-02-27
Extra capi v1beta2 stuff

### LINKED ISSUE #96: Too many calls when deleting machine

wikkyk:
*Description of changes:*
Just some additional changes to #639 that I think would be nice. Each commit is reasonably self-contained and most can be cherry-picked as they are.

1. 705e360 controversial but it doesn't remove the markers from the CAPI-mandated fields, afaict. we need to discuss this further.
2. 3632a31 adds port validation to APIEndpoint (without it, validation never fires on missing port), resurrects IsZero() and String() (I think these are useful).
3. 9a29fc7 if ProviderID is optional then it should be a pointer. also adds proper uuid validation. If we don't want ProviderID to be a pointer, then the extra validation is still useful.
4. 741c73c big change that keeps round-trip compat with old failure stuff. the compat fields are in dedicated structs, easy to remove for v1alpha3. depends on IsZero() and String() from 3632a31, resurrects HasFailed().
5. 8abf97f refactors deletion slightly. I noticed a rare bug (race?) where deletion cannot proceed due to missing credentials (removed earlier?) we've seen some deletion-related issues before. maybe this finally fixes #96. depends on a constant introduced in 741c73c.
6. 1b5e197 completes ipam v1beta2
7. d9f5b8d fluff

I like the helpers that I re-introduced because they neatly encapsulate the semantics and result in shorter, clearer code at the call-sites. It is also code that keeps repeating (DRY). If we ever need to perform more work to accomplish what the helper does, we only have 1 place to change. If it ever has a bug, there is only 1 place to change.

*Testing performed:*
`make test`, some tests have been changed to reflect new behaviour and some tests have been added

### DIFF
--- a/.golangci-kal.yml
+++ b/.golangci-kal.yml
@@ -18,7 +18,6 @@ linters:
               - "conditions"
               - "conflictingmarkers"
               - "duplicatemarkers"
-              - "forbiddenmarkers"
               - "integers"
               - "jsontags"
               - "nobools"


wikkyk:
I left the KAL failures because we're still discussing how to handle defaults.

wikkyk:
as @mcbenjemaa pointed out off-PR, the deletion commit doesn't make sense because without a scope you can't do anything anyway.
I removed the deprecated fields addition as discussed off-PR, we don't need it. The commit is replaced with a smaller commit that has a few improvements.

api/v1alpha1/proxmoxclustertemplate_conversion.go
```diff
@@ -84,6 +84,13 @@ func (src *ProxmoxClusterTemplate) ConvertTo(dstRaw conversion.Hub) error {
 
 	}
 
+	// ObjectMeta.Annotations is copied by reference from the spoke during
+	// auto-conversion. After UnmarshalData removes the conversion-data key,
+	// we may be left with an empty map instead of nil.
+	if len(dst.Annotations) == 0 {
```
mcbenjemaa: This shouldn't be here.
I spent hours getting it out of here.

api/v1alpha1/proxmoxmachine_conversion.go
```diff
@@ -44,13 +44,21 @@ func (src *ProxmoxMachine) ConvertTo(dstRaw conversion.Hub) error {
...
 		dst.Status.VMStatus = nil
 	}
 
 	// Normalize ProxmoxMachineSpec after auto-conversion
 	normalizeProxmoxMachineSpec(&dst.Spec)
 
+	// ObjectMeta.Annotations is copied by reference from the spoke during
+	// auto-conversion. After UnmarshalData removes the conversion-data key,
+	// we may be left with an empty map instead of nil.
+	if len(dst.Annotations) == 0 {
```
mcbenjemaa: This one also shouldn't be here,
these are already normilized in the fuzzerfuncs

api/v1alpha1/proxmoxmachine_conversion.go
```diff
@@ -44,13 +44,21 @@ func (src *ProxmoxMachine) ConvertTo(dstRaw conversion.Hub) error {
...
+		dst.Annotations = nil
```
mcbenjemaa: Why do I set annotations to nil after conversion?

api/v1alpha1/proxmoxmachinetemplate_conversion.go
```diff
@@ -41,6 +41,13 @@ func (src *ProxmoxMachineTemplate) ConvertTo(dstRaw conversion.Hub) error {
 	// Normalize ProxmoxMachineSpec after auto-conversion
 	normalizeProxmoxMachineSpec(&dst.Spec.Template.Spec)
 
+	// ObjectMeta.Annotations is copied by reference from the spoke during
+	// auto-conversion. After UnmarshalData removes the conversion-data key,
+	// we may be left with an empty map instead of nil.
+	if len(dst.Annotations) == 0 {
```
mcbenjemaa: same remark

api/v1alpha1/proxmoxmachinetemplate_conversion.go
```diff
@@ -41,6 +41,13 @@ func (src *ProxmoxMachineTemplate) ConvertTo(dstRaw conversion.Hub) error {
...
+		dst.Annotations = nil
```
mcbenjemaa: nil ?

api/v1alpha2/proxmoxcluster_types.go
```diff
@@ -244,6 +246,7 @@ type ProxmoxClusterStatus struct {
 	// initialization provides observations of the ProxmoxCluster initialization process.
 	// NOTE: Fields in this struct are part of the Cluster API contract and are used to orchestrate initial Cluster provisioning.
 	// +optional
+	// +kubebuilder:default={provisioned: false}
```
mcbenjemaa: This is unnecessary and meaningless. 
provisioned is a pointer.

api/v1alpha2/proxmoxmachine_types.go
```diff
@@ -457,6 +457,7 @@ type ProxmoxMachineStatus struct {
 	// initialization provides observations of the ProxmoxMachine initialization process.
 	// NOTE: Fields in this struct are part of the Cluster API contract and are used to orchestrate initial Machine provisioning.
 	// +optional
+	// +kubebuilder:default={provisioned: false}
```
mcbenjemaa: Same, no need for defaulting this.

wikkyk:
dropped the ProviderID patch as it is merged into capi/v1beta2

api/v1alpha1/proxmoxclustertemplate_conversion.go
```diff
@@ -84,6 +84,13 @@ func (src *ProxmoxClusterTemplate) ConvertTo(dstRaw conversion.Hub) error {
...
+		dst.Annotations = nil
```
mcbenjemaa: nil annotations after conversion

internal/controller/proxmoxcluster_controller.go
```diff
@@ -192,7 +190,26 @@ func (r *ProxmoxClusterReconciler) reconcileNormal(ctx context.Context, clusterS
 	// If the ProxmoxCluster doesn't have our finalizer, add it.
 	ctrlutil.AddFinalizer(clusterScope.ProxmoxCluster, infrav1.ClusterFinalizer)
 
+	// Ensure initialization status is always set so the status subresource
+	// passes CRD validation (initialization is required with minProperties:1).
+	// SetReady() will override this to true when all reconciliation succeeds.
+	if clusterScope.ProxmoxCluster.Status.Initialization.Provisioned == nil {
```
mcbenjemaa: This statement is also unnecessary.

internal/controller/proxmoxcluster_controller.go
```diff
@@ -192,7 +190,26 @@ func (r *ProxmoxClusterReconciler) reconcileNormal(ctx context.Context, clusterS
...
+		clusterScope.ProxmoxCluster.Status.Initialization.Provisioned = ptr.To(false)
+	}
+
 	if ptr.Deref(clusterScope.ProxmoxCluster.Spec.ExternalManagedControlPlane, false) {
+		if clusterScope.ProxmoxCluster.Spec.ControlPlaneEndpoint.IsZero() {
```
mcbenjemaa, line 210: This condition does not make sense, while `ExternalManagedControlPlane` is true.
It will prevent the cluster with an externally managed control plane, like Kamaji, from reconciling.

internal/service/vmservice/bootstrap.go
```diff
@@ -241,7 +241,12 @@ func getNetworkConfigDataForDevice(ctx context.Context, machineScope *scope.Mach
 	ipConfigs := make([]types.IPConfig, 0, len(ipAddresses))
 	for _, ipAddr := range ipAddresses {
 		ipConfig := types.IPConfig{}
-		ip, err := netip.ParsePrefix(fmt.Sprintf("%s/%d", ipAddr.Spec.Address, ptr.Deref(ipAddr.Spec.Prefix, 0)))
+		// In v1beta2, Prefix is a *int32 pointer, so we need to dereference it
+		prefix := int32(24) // default prefix if not specified
+		if ipAddr.Spec.Prefix != nil {
+			prefix = *ipAddr.Spec.Prefix
+		}
```
mcbenjemaa: This block is reinventing the wheel and does exactly what that single line above does with`ptr.Deref`.

internal/service/vmservice/helpers_test.go
```diff
@@ -235,10 +235,16 @@ func setupReconcilerTest(t *testing.T) (*scope.MachineScope, *proxmoxtest.MockCl
 }
 
 func createIPAddressResource(t *testing.T, c client.Client, name string, machineScope *scope.MachineScope, ip netip.Prefix, offset int, pool *corev1.TypedLocalObjectReference) {
-	prefix := ip.Bits()
+	prefix := int32(ip.Bits())
 	var gateway string
+	var poolRef ipamv1.IPPoolReference
```
mcbenjemaa, line 240: I don't understand the need for this variable?

internal/service/vmservice/helpers_test.go
```diff
@@ -257,30 +263,18 @@ func createIPAddressResource(t *testing.T, c client.Client, name string, machine
 				}},
 			},
 			Spec: ipamv1.IPAddressClaimSpec{
-				PoolRef: ipamv1.IPPoolReference{
-					APIGroup: ptr.Deref(pool.APIGroup, ""),
-					Kind:     pool.Kind,
-					Name:     pool.Name,
-				},
+				PoolRef: poolRef,
```
mcbenjemaa, line 266: The changes introduced does the same thing, I don't see any improvements.

pkg/kubernetes/ipam/ipam.go
```diff
@@ -497,11 +515,11 @@ func (h *Helper) CreateIPAddressClaim(ctx context.Context, owner client.Object,
 			Annotations: annotations,
 		},
 		Spec: ipamv1.IPAddressClaimSpec{
-			PoolRef: ipamv1.IPPoolReference{
-				APIGroup: gvk.Group,
+			PoolRef: typedLocalObjectRefToIPPoolRef(corev1.TypedLocalObjectReference{
```
mcbenjemaa: I don't see any value for this function.
The current code looks good.

pkg/kubernetes/ipam/ipam.go
```diff
@@ -90,6 +90,24 @@ func isIPv4(ip string) (bool, error) {
...
+func ipPoolRefToTypedLocalObjectRef(ref ipamv1.IPPoolReference) corev1.TypedLocalObjectReference {
+	return corev1.TypedLocalObjectReference{
+		APIGroup: ptr.To(ref.APIGroup),
+		Kind:     ref.Kind,
+		Name:     ref.Name,
+	}
+}
+
+// typedLocalObjectRefToIPPoolRef converts a corev1.TypedLocalObjectReference to ipamv1.IPPoolReference.
+func typedLocalObjectRefToIPPoolRef(ref corev1.TypedLocalObjectReference) ipamv1.IPPoolReference {
```
mcbenjemaa: check the comment related to this function,

pkg/kubernetes/ipam/ipam.go
```diff
@@ -90,6 +90,24 @@ func isIPv4(ip string) (bool, error) {
 	return netIP.Is4(), nil
 }
 
+// ipPoolRefToTypedLocalObjectRef converts an ipamv1.IPPoolReference to corev1.TypedLocalObjectReference.
+func ipPoolRefToTypedLocalObjectRef(ref ipamv1.IPPoolReference) corev1.TypedLocalObjectReference {
```
mcbenjemaa: This is also the same function as the defined function, `toTypedLocalObjectReference`.
Rename it if you don't like the function name. I'm Ok with that.

internal/service/vmservice/helpers_test.go
wikkyk: It's shorter. The old code created two identical IPPoolReferences and had two identical subsequent ifs.

internal/service/vmservice/helpers_test.go
wikkyk: It's meant to do the same thing but it's shorter.

config/crd/bases/infrastructure.cluster.x-k8s.io_proxmoxmachinetemplates.yaml
```diff
@@ -1182,19 +1160,17 @@ spec:
...
                         description: pool adds the new VM to the specified pool.
-                        maxLength: 256
                         type: string
                       providerID:
                         description: |-
                           providerID is the virtual machine BIOS UUID formatted as
                           proxmox://6c3fa683-bef9-4425-b413-eaa45a9d6191
                         maxLength: 46
-                        minLength: 46
+                        minLength: 1
```
wikkyk: this is a stray change but I'll regen after merging anyway

config/crd/bases/infrastructure.cluster.x-k8s.io_proxmoxmachines.yaml
```diff
@@ -1307,19 +1285,17 @@ spec:
...
                 description: pool adds the new VM to the specified pool.
-                maxLength: 256
                 type: string
               providerID:
                 description: |-
                   providerID is the virtual machine BIOS UUID formatted as
                   proxmox://6c3fa683-bef9-4425-b413-eaa45a9d6191
                 maxLength: 46
-                minLength: 46
+                minLength: 1
```
wikkyk: stray, will regen

api/v1alpha2/proxmoxcluster_types.go
```diff
@@ -44,15 +46,14 @@ type ProxmoxClusterSpec struct {
 	// externalManagedControlPlane can be enabled to allow externally managed Control Planes to patch the
 	// Proxmox cluster with the Load Balancer IP provided by Control Plane provider.
 	// +optional
+	// +default=false
 	ExternalManagedControlPlane *bool `json:"externalManagedControlPlane,omitempty"`
```
mcbenjemaa, line 50: you may want to disable forbiddenmarkers default from KAL

internal/controller/proxmoxcluster_controller.go
mcbenjemaa: Ok, got the condition,

You can delete all 3 conditions below and just keep this one.

internal/service/vmservice/vm.go
```diff
@@ -318,8 +321,8 @@ func reconcileVirtualMachineConfig(ctx context.Context, machineScope *scope.Mach
 		devices := machineScope.ProxmoxMachine.Spec.Network.NetworkDevices
 		for _, v := range devices {
 			vmOptions = append(vmOptions, proxmox.VirtualMachineOption{
-				Name:  ptr.Deref(v.Name, infrav1.DefaultNetworkDevice),
-				Value: formatNetworkDevice(ptr.Deref(v.Model, "virtio"), ptr.Deref(v.Bridge, ""), v.MTU, v.VLAN),
+				Name:  *v.Name,
+				Value: formatNetworkDevice(*v.Model, *v.Bridge, v.MTU, v.VLAN),
```
mcbenjemaa, line 325: sgh, why this has been changed like this:

mcbenjemaa: This can cause a nil pointer dereference.

internal/service/vmservice/utils.go
```diff
@@ -158,7 +158,7 @@ func shouldUpdateNetworkDevices(machineScope *scope.MachineScope) bool {
 		bridge := extractNetworkBridge(net)
 
 		// current is different from the desired spec.
-		if model != ptr.Deref(v.Model, "") || bridge != ptr.Deref(v.Bridge, "") {
+		if model != *v.Model || bridge != *v.Bridge {
```
mcbenjemaa, line 161: nil pointer dereference.

wikkyk:
Closing because I merged the patches manually, besides the deletion changes.

PR #655, opened 2026-03-04
V1alpha2/fixes

wikkyk:
*Description of changes:*
A few minor fixes for issues brought up by SonarQube.

*Testing performed:*
`make test`

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -149,7 +149,7 @@ func main() {
 		os.Exit(1)
 	}
 
-	if setupErr := setupReconcilers(ctx, mgr, pmoxClient); setupErr != nil {
+	if err := setupReconcilers(ctx, mgr, pmoxClient); err != nil {
 		setupLog.Error(err, "unable to setup reconcilers")
 		os.Exit(1)
 	}


cmd/main.go
```diff
@@ -249,6 +237,22 @@ func setupProxmoxClient(ctx context.Context, logger logr.Logger) (capmox.Client,
 	)
 }
 
+func setupWebhooks(mgr ctrl.Manager) (string, error) {
+	if err := (&webhook.ProxmoxCluster{}).SetupWebhookWithManager(mgr); err != nil {
+		return "ProxmoxCluster", err
```
mcbenjemaa: Why not return just the formatted error with the type.

cmd/main.go
```diff
@@ -166,20 +166,8 @@ func main() {
...
-			setupLog.Error(err, "unable to create webhook", "webhook", "ProxmoxMachine")
-			os.Exit(1)
-		}
-		if err = (&webhook.ProxmoxMachineTemplate{}).SetupWebhookWithManager(mgr); err != nil {
-			setupLog.Error(err, "unable to create webhook", "webhook", "ProxmoxMachineTemplate")
-			os.Exit(1)
-		}
-		if err = (&webhook.ProxmoxClusterTemplate{}).SetupWebhookWithManager(mgr); err != nil {
-			setupLog.Error(err, "unable to create webhook", "webhook", "ProxmoxClusterTemplate")
+		if what, err := setupWebhooks(mgr); err != nil {
```
mcbenjemaa: err is enough for the return.

cmd/main.go
wikkyk: Just to follow the pattern. I didn't want to change the output or behaviour here. Mostly just to remove the repeated string constant.

cmd/main.go
wikkyk: how?

mcbenjemaa: In the setupWebhooks function, you can use `errors.Wrapf()`
like:
```
errors.Wrapf(err, "unable to create webhook %s", "ProxmoxCluster")
```

So, the log here becomes,

```
setupLog.Error(err, "unable to setup webhooks")
```

and as you can see, this is the same pattern we use in `setupReconcilers`

65278: I agree, WrapF the inner error. Don't return a string (that'll just trigger complaints about translating to constants anyway).

wikkyk: This actually removes the complaint about duplicated strings.

cmd/main.go
```diff
@@ -166,20 +166,8 @@ func main() {
...
-			os.Exit(1)
-		}
-		if err = (&webhook.ProxmoxMachineTemplate{}).SetupWebhookWithManager(mgr); err != nil {
-			setupLog.Error(err, "unable to create webhook", "webhook", "ProxmoxMachineTemplate")
-			os.Exit(1)
-		}
-		if err = (&webhook.ProxmoxClusterTemplate{}).SetupWebhookWithManager(mgr); err != nil {
-			setupLog.Error(err, "unable to create webhook", "webhook", "ProxmoxClusterTemplate")
+		if err := setupWebhooks(mgr); err != nil {
+			setupLog.Error(err, "unable to create webhook")
```
mcbenjemaa:
```suggestion
			setupLog.Error(err, "unable to setup webhooks")
```

PR #656, opened 2026-03-04
.github: remove comment workflow

wikkyk:
It doesn't really add give any useful information. GitHub UI already tells the contributor that actions are awaiting approval. Choosing the correct e2e is the responsibility of the codeowner approving the run.


wikkyk:
Good point. Docs updated.

Issue #635, opened 2026-02-17
Maintenance/version bump helpers


wikkyk:
Most dependency bumps are straightforward but there are a few that require modifying several files (and that's just to apply the bump without any porting work).

Examples include:
* versions that are referenced in multiple places should be synchronised: go versions and required packages in both go.mods, linters in go.mod and action yaml, k8s.io/* and envtest in go.mod, Makefile, docs, possibly controller-runtime and controller-tools
* multi-part packages, e.g cluster-api{,-test} should be the same version, ginkgo/gomega are best updated at the same time
* cluster-api's version is pinned with replace, dependabot only rewrites the version under require and leaves the replace intact
* versions referred to in documentation, examples, test cases, e2e, ...

Bumping these is always a tedious chore because dependabot produces incomplete PRs that only bump one go.mod, only bump the action, skips versions etc. Incomplete automation adds work and most of this is pure churn.

We should come up with at least a linter that checks if versions of these special deps are consistent everywhere they are referred.
The ideal solution would also have a script, or a bunch of scripts, that bump versions. These could be executed manually (some bumps e.g. capi and k8s would typically be done manually because they tend to require porting and docs updates) and/or in an action (so that e.g. a dependabot bump gets replicated in the other go.mod/action in the same PR).

PR #657, opened 2026-03-04
convert lint to commit hashes, replace curlbash with action

wikkyk:
*Description of changes:*

We should really start converting `@version` with `@commithash #version` in action. This PR does the lint action.
While I was at it, I replaced the actionlint curlbash with a proper action.

*Testing performed:*
run the actions

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -25,6 +25,10 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v6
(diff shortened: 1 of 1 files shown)


wikkyk:
SonarQube complained about the workflow-level permissions but in this case all linters should be able to write annotations. The TODO is for when we get around to #635.

wikkyk:
@copilot update/rebase this PR on top of the changes to the lint workflow in PR #672.

PR #671, opened 2026-03-09
Move tool management from Makefile to go.mod tool

wikkyk:
*Description of changes:*
Moved the tools installed from Makefile to go.mod/tool and converted invocations to `go tool`.
Simplified the tool logic because the tools no longer need an explicit fetch/install step - `go tool` will install the tools if they're not installed.
Version pins are lifted from the Makefile but I'm not sure if we actually need them all, e.g. envsubst, yamlfmt, mockery, kustomize.
I also updated kustomize to the latest v5 mainly due to compat issues with v5.0.0.

*Testing performed:*
ran the programs using make targets.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,16 +3,12 @@ module github.com/ionos-cloud/cluster-api-provider-proxmox
 go 1.25.0
 
 replace (
-	github.com/a8m/envsubst => github.com/a8m/envsubst v1.4.2
 	github.com/golangci/golangci-lint/v2 => github.com/golangci/golangci-lint/v2 v2.9.0
-	github.com/google/yamlfmt => github.com/google/yamlfmt v0.21.0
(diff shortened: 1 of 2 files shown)


wikkyk:
In the last commit, I removed the following version pins:

- envsubst (pin was 1.4.2, upstream is at 1.4.3, gives the same output)
- yamlfmt (pin was at upstream anyway, it's not a critical dep anyway)
- mockery/v2 (pin was at upstream, v2 shouldn't change behaviour - if it does and we don't have the resources to immediately port, we can always restore the pin)
- kustomize/v5 (I had to update it from v5.0.0 anyway due to incompatibility with github.com/google/gnostic{,-models}/openapiv2, upstream v5.8.1 works fine and v5 should continue to produce the same output)

lmk if any of this is controversial.

wikkyk:
I'm not sure if sigs.k8s.io/controller-tools and -runtime shouldn't be pinned to the same version.

PR #675, opened 2026-03-10
v1alpha2: turn infrav2.NetName back into a string

65278:
*Issue #, if available:*

*Description of changes:*
v1alpha2.NetName was a pointer in my prototype, and when Mohamed made the conversion hook, he requested no pointer changes because of rebasing.

I'm slipping this one in at the last moment, because it really shouldn't be a pointer. Name is a required field in NetworkDevices and can never be nil.

*Testing performed:*
Deployed a ClusterClass with it. Make test finishes. Linting performed.

### DIFF
--- a/api/v1alpha2/proxmoxmachine_types.go
+++ b/api/v1alpha2/proxmoxmachine_types.go
@@ -39,7 +39,7 @@ const (
 	DefaultReconcilerRequeue = 10 * time.Second
 
 	// DefaultNetworkDevice is the default network device name.
-	DefaultNetworkDevice = "net0"
+	DefaultNetworkDevice = NetName("net0")
 
 	// DefaultSuffix is the default suffix for the network device.
 	DefaultSuffix = "inet"
@@ -514,8 +514,8 @@ type ProxmoxMachineInitializationStatus struct {
 // IPAddressesSpec stores the IP addresses of a network interface. Used for status.
 type IPAddressesSpec struct {
 	// net is the proxmox network name these ipaddresses are attached to.
+	// +kubebuilder:validation:MinLength=4
 	// +kubebuilder:validation:Pattern=`^(net[0-9]+|default)$`
-	// +kubebuilder:validation:MinLength=1
 	// +required
 	NetName string `json:"net,omitempty"`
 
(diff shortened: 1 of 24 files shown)


api/v1alpha2/proxmoxmachine_types.go
```diff
@@ -514,8 +514,8 @@ type ProxmoxMachineInitializationStatus struct {
 // IPAddressesSpec stores the IP addresses of a network interface. Used for status.
 type IPAddressesSpec struct {
 	// net is the proxmox network name these ipaddresses are attached to.
+	// +kubebuilder:validation:MinLength=4
```
wikkyk, line 517: nice

api/v1alpha2/proxmoxmachine_types.go
```diff
@@ -514,8 +514,8 @@ type ProxmoxMachineInitializationStatus struct {
...
 	// +kubebuilder:validation:Pattern=`^(net[0-9]+|default)$`
-	// +kubebuilder:validation:MinLength=1
 	// +required
 	NetName string `json:"net,omitempty"`
```
wikkyk, line 520: shouldn't the type be NetName here? wouldn't need to double-up with the validation.

internal/service/vmservice/utils.go
wikkyk, line 146: perhaps it's too late to do this now and we should open a todo issue about it but nets could be a map[NetName]... rather than map[string]... . I would even go as far as to say that this applies to all places where you cast between NetName and string - every func that takes or returns a network name and every data structure that contains one should always use NetName and never bare string.

api/v1alpha2/proxmoxmachine_types.go
65278: No, shouldn't be. This can be "default". a Proxmox interface name can never be default.

internal/service/vmservice/utils.go
65278: Can't be, we're not in charge of VirtualMachineConfig.MergeNets(), that's goproxmox.

api/v1alpha2/proxmoxmachine_types.go
65278: o/

internal/service/vmservice/utils.go
65278: I also can't avoid strings for v1alpha1 in conversion. I've done as many as I can.
There's one open case where the network config datastructure compares a placeholder name to NetName where a cast is required. I'll have to rethink that.

PR #679, opened 2026-03-12
deps: bump k8s to 33, controller-tools to 18, controller-runtime to 21

wikkyk:
*Description of changes:*
Does the bumps from v1alpha2/wip, except for CAPI.

These versions match the versions in v1alpha2/wip.
CAPI omitted because 1.11 is implemented by v1alpha2 and so requires the extensive code changes.

I used the bump-k8s tool from #669 to do this then tweaked slightly to align the other deps.

*Testing performed:*
make test
It doesn't need to be perfect, I'm just staging changes for the 0.8.0 release.


wikkyk:
the e2e failure is irrelevant, we need to fix that in v1alpha2 not here.

PR #680, opened 2026-03-16
Template selection by tag subset

65278:
*Issue #, if available:*
#583 

*Description of changes:*
This is a port of #584 to v1alpha2. I've added another matching policy `BestSubset` which finds the best template distance between your tags and the template tags. It still errors if you get more than one `best` match.

*Testing performed:*
All changes tested on a real proxmox cluster, except for BestSubset policy.


wikkyk:
Do we need the v1alpha1 changes?

wikkyk:
I squashed the commits by logic and updated copyright headers where necessary.

wikkyk:
The code duplication is literally just the test being in both v1alpha1 and v1alpha2. Safe to ignore.

Issue #714, opened 2026-04-10
[v0.8.0] ProxmoxMachine.status.addresses not populated after v1alpha1→v1alpha2 migration, breaking RKE2 bootstrap (internal-first)


ibra-coul:
**What steps did you take and what happened:**
After migrating from infrastructure.cluster.x-k8s.io/v1alpha1 to v1alpha2, existing ProxmoxMachine objects no longer have status.addresses populated. As a result, Machine.status.addresses is also null, causing the RKE2 bootstrap controller to loop indefinitely on:

"No ControlPlane IP Address found for node registration"
This blocks every rolling update step — new VMs are never created in Proxmox because bootstrap data is never generated.

**What did you expect to happen:**
After migration to v1alpha2, CAPMOX should re-reconcile existing ProxmoxMachine objects and repopulate status.addresses automatically.

**Environment:**

- Proxmox version:  9.1.1
- Cluster-api-provider-proxmox version: v1.12.5
- Kubernetes version: (use `kubectl version`): v1.34.4+rke2r1
- OS (e.g. from `/etc/os-release`): Ubuntu 24.04.4 LTS
- CAPMOX version: v0.8.0 (ghcr.io/ionos-cloud/cluster-api-provider-proxmox:v0.8.0)
- CAPI version: v1beta2
- Bootstrap provider: RKE2 with registrationMethod: internal-first
- Trigger: migration of existing cluster from v1alpha1 → v1alpha2

**Steps to reproduce**
- Have a running cluster provisioned with CAPMOX v1alpha1
- Upgrade CAPMOX to v0.8.0 (which adds v1alpha2 support)
- Update manifests to use infrastructure.cluster.x-k8s.io/v1alpha2
- Trigger a rolling update (e.g. change a ProxmoxMachineTemplate field)
- Observe ProxmoxMachine.status.addresses is null on existing machines

```bash
kubectl get proxmoxmachine -n <namespace> -o json | \
  jq '.items[] | {name: .metadata.name, addresses: .status.addresses}'
# → all null
```

**Root cause**
CAPMOX does not re-reconcile existing ProxmoxMachine objects after the API version migration. The status.addresses field is never populated unless a reconciliation is explicitly triggered

**Workaround**
Manually annotate each ProxmoxMachine to force reconciliation, then patch Machine.status.addresses:

```bash
kubectl annotate proxmoxmachine <name> -n <namespace> \
[...]

Issue #710, opened 2026-04-09
v0.8.0: ProxmoxMachine.status.addresses never populated; CAPI Machine has no addresses (Talos CP blocks on EtcdClusterHealthy)


Leonorus:
**What steps did you take and what happened:**

Upgraded from capmox v0.7.7 → v0.8.0 (v1alpha2 / CAPI v1beta2). Freshly created `ProxmoxMachine` objects never get `.status.addresses` populated. `.status.ipAddresses` (the custom field) is filled correctly from IPAM, but the standard CAPI `[]MachineAddress` list stays nil. The CAPI `Machine` therefore also ends up with empty `.status.addresses`.

For Talos-based clusters (sidero `cluster-api-control-plane-provider-talos`), this blocks the control plane forever:

```
ControlPlaneComponentsHealthy=False ControlPlaneComponentsInspectionFailed
  machine "qak8scapicl1-cp-cz4vt": no addresses were found for node "qak8scapicl1-cp-cz4vt"
EtcdClusterHealthyCondition=False EtcdClusterUnhealthy
  error checking etcd health on machine "qak8scapicl1-cp-cz4vt": no addresses were found for node ...
```

The VMs are cloned, powered on, reachable on the network, Talos is up and etcd is running — but the Talos control-plane provider has no addresses to reach them on, so health checks never run and the cluster stays `Available=False` forever.

Observed `ProxmoxMachine` state:

```yaml
status:
  conditions:
  - type: Ready
    status: "True"
    reason: InfoReported
  - type: VirtualMachineProvisioned
    status: "True"
    reason: Provisioned
  ipAddresses:
  - ipv4: ["192.168.160.3"]
    net: default
  - ipv4: ["192.168.160.3"]
    net: net0
  vmStatus: ready
  bootstrapDataProvided: true
  # addresses: NOT SET
```

Restarting `capmox-controller-manager` does not re-populate the field on subsequent reconciles.

**What did you expect to happen:**

`ProxmoxMachine.status.addresses` should be populated with a `Hostname` + one or more `InternalIP` entries from the default (`net0`) network device, the same way v0.7 did, so that CAPI and any control-plane provider (Kubeadm, Talos, …) can resolve the node.

**Anything else you would like to add:**

[...]

PR #721, opened 2026-04-13
fix(vm_service): state machine does not run reconcileMachineAddresses

65278:
*Issue #, if available:*
#710
#714 

*Description of changes:*
The state machine was set to the wrong state after reconcilePowerState, therefore proxmoxmachine.Status.Addresses was never populated. Our tests for reconcileMachineAddresses succesfully ran as the code itself was fine. Our E2E tests did not fail because we do not depend on this field.

To catch these kinds of errors in the future, I've made a new test for ReconcileVM which tries to test its workings in their entirety.

*Testing performed:*
`make test`

### DIFF
--- a/internal/service/taskservice/task_test.go
+++ b/internal/service/taskservice/task_test.go
@@ -123,7 +123,7 @@ func setupTaskTest(t *testing.T) (*scope.MachineScope, *proxmoxtest.MockClient) 
 }
 
 // Do not return task if there is no task.
-func TestGetTask_NilTaskRef(t *testing.T) {
+func TestGetTask_NoTaskRef(t *testing.T) {
 	machineScope, _ := setupTaskTest(t)
 	machineScope.ProxmoxMachine.Status.TaskRef = nil
 
@@ -133,7 +133,7 @@ func TestGetTask_NilTaskRef(t *testing.T) {
 }
 
 // Test missing task erroring.
-func TestGetTask_Error(t *testing.T) {
+func TestGetTask_ErrorNotFound(t *testing.T) {
 	machineScope, mockClient := setupTaskTest(t)
 	machineScope.ProxmoxMachine.Status.TaskRef = ptr.To("UPID:node1:001")
 
@@ -158,7 +158,7 @@ func TestGetTask_Success(t *testing.T) {
 }
 
 // Test ReconcileInFlightTask on empty task.
-func TestReconcileInFlightTask_NilTaskRef(t *testing.T) {
+func TestReconcileInFlightTask_NoTaskRef(t *testing.T) {
 	machineScope, _ := setupTaskTest(t)
 	machineScope.ProxmoxMachine.Status.TaskRef = nil
 
@@ -168,7 +168,7 @@ func TestReconcileInFlightTask_NilTaskRef(t *testing.T) {
 }
 
 // Test ReconcileInFlightTask on empty task but existing TaskRef.
-func TestReconcileInFlightTask_NilTask(t *testing.T) {
+func TestReconcileInFlightTask_NilTaskReturned(t *testing.T) {
 	machineScope, mockClient := setupTaskTest(t)
 	machineScope.ProxmoxMachine.Status.TaskRef = ptr.To("UPID:node1:001")
 
@@ -207,7 +207,7 @@ func TestReconcileInFlightTask_TaskSuccessful(t *testing.T) {
 }
 
 // Test ReconcileInflightTask on task failure switch case if not qmstart.
-func TestReconcileInFlightTask_TaskFailed_NonQMStart(t *testing.T) {
+func TestReconcileInFlightTask_CloneTaskFailed(t *testing.T) {
 	machineScope, mockClient := setupTaskTest(t)
 	machineScope.ProxmoxMachine.Status.TaskRef = ptr.To("UPID:node1:001")
 
--- a/internal/service/taskservice/task_test.go
+++ b/internal/service/taskservice/task_test.go
@@ -1,0 +1,332 @@
+/*
+Copyright 2023-2026 IONOS Cloud.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package taskservice
+
+import (
+	"context"
+	"errors"
+	"testing"
+	"time"
+
+	"github.com/go-logr/logr"
+	"github.com/luthermonson/go-proxmox"
+	"github.com/stretchr/testify/require"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/utils/ptr"
+	clusterv1 "sigs.k8s.io/cluster-api/api/core/v1beta2"
+	"sigs.k8s.io/cluster-api/util/conditions"
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
+
+	infrav1 "github.com/ionos-cloud/cluster-api-provider-proxmox/api/v1alpha2"
+	"github.com/ionos-cloud/cluster-api-provider-proxmox/pkg/kubernetes/ipam"
+	"github.com/ionos-cloud/cluster-api-provider-proxmox/pkg/proxmox/proxmoxtest"
+	"github.com/ionos-cloud/cluster-api-provider-proxmox/pkg/scope"
+)
+
+func setupTaskTest(t *testing.T) (*scope.MachineScope, *proxmoxtest.MockClient) {
+	t.Helper()
+
+	cluster := &clusterv1.Cluster{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "test",
+			Namespace: metav1.NamespaceDefault,
+		},
+	}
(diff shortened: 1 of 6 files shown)


internal/service/vmservice/vm_test.go
```diff
@@ -695,3 +696,163 @@ func TestReconcileVM_CloudInitRunning(t *testing.T) {
...
+		},
+	)
+
+	proxmoxClient.EXPECT().GetVM(context.Background(), "node1", int64(123)).Return(vm, nil).Once()
+	proxmoxClient.EXPECT().StartVM(context.Background(), vm).Return(newTask(), nil).Once()
+
+	// Provide IPAddresses fields to fake network bootstrap.
+	machineScope.ProxmoxMachine.Status.IPAddresses = []infrav1.IPAddressesSpec{{
+		NetName: string(infrav1.DefaultNetworkDevice),
+		IPv4:    []string{"10.10.10.10"},
```
wikkyk: #693 please

65278: Done

65278:
I've added a CloneVM mock to the testcase, now it's rather complete. The only thing that could be added would be a full bootstrap.go test, but that's a full can of worms I'll leave to somebody else to implement.

Now we know that you need to requeue at least 4 times to deploy a capmox machine.

65278:
Actual deployment performed. It turns out we had problems way deeper than the initial problem:

A qmstart task always fails on our proxmox. The VM starts anyway with two different errors. I've made it so that taskfailure now knows about qmstart and treats it as a special case.
The failure in tasks caused the state machine to always run incorrectly even after the naive fix. This is also fixed now.

Because this triggered the quality gate, I've had claude add a test for task.go, even though we're only testing invariants and the mock kubernetes client.

internal/service/taskservice/task_test.go
```diff
@@ -0,0 +1,332 @@
...
+	machineScope, _ := setupTaskTest(t)
+	machineScope.ProxmoxMachine.Status.TaskRef = nil
+
+	task, err := GetTask(context.Background(), machineScope)
+	require.NoError(t, err)
+	require.Nil(t, task)
+}
+
+// Test missing task erroring.
+func TestGetTask_Error(t *testing.T) {
```
mcbenjemaa:
```suggestion
func TestGetTask_ErrorNotFound(t *testing.T) {
```

internal/service/taskservice/task_test.go
```diff
@@ -0,0 +1,332 @@
...
+		ProxmoxMachine: infraMachine,
+		IPAMHelper:     ipamHelper,
+	})
+	require.NoError(t, err)
+
+	return machineScope, mockClient
+}
+
+// Do not return task if there is no task.
+func TestGetTask_NilTaskRef(t *testing.T) {
```
mcbenjemaa:
```suggestion
func TestGetTask_NoTaskRef(t *testing.T) {
```

internal/service/taskservice/task_test.go
```diff
@@ -0,0 +1,332 @@
...
+	expected := &proxmox.Task{UPID: "UPID:node1:001", IsCompleted: true, IsSuccessful: true}
+	mockClient.EXPECT().GetTask(context.Background(), "UPID:node1:001").Return(expected, nil).Once()
+
+	task, err := GetTask(context.Background(), machineScope)
+	require.NoError(t, err)
+	require.Equal(t, expected, task)
+}
+
+// Test ReconcileInFlightTask on empty task.
+func TestReconcileInFlightTask_NilTaskRef(t *testing.T) {
```
mcbenjemaa:
```suggestion
func TestReconcileInFlightTask_NoTaskRef(t *testing.T) {
```

internal/service/taskservice/task_test.go
```diff
@@ -0,0 +1,332 @@
...
+	machineScope, _ := setupTaskTest(t)
+	machineScope.ProxmoxMachine.Status.TaskRef = nil
+
+	requeue, err := ReconcileInFlightTask(context.Background(), machineScope)
+	require.NoError(t, err)
+	require.False(t, requeue)
+}
+
+// Test ReconcileInFlightTask on empty task but existing TaskRef.
+func TestReconcileInFlightTask_NilTask(t *testing.T) {
```
mcbenjemaa:
```suggestion
func TestReconcileInFlightTask_NilTaskReturned(t *testing.T) {
```

internal/service/taskservice/task_test.go
```diff
@@ -0,0 +1,332 @@
...
+	mockClient.EXPECT().GetTask(context.Background(), "UPID:node1:001").Return(task, nil).Once()
+
+	requeue, err := ReconcileInFlightTask(context.Background(), machineScope)
+	require.NoError(t, err)
+	require.False(t, requeue)
+	require.Nil(t, machineScope.ProxmoxMachine.Status.TaskRef)
+}
+
+// Test ReconcileInflightTask on task failure switch case if not qmstart.
+func TestReconcileInFlightTask_TaskFailed_NonQMStart(t *testing.T) {
```
mcbenjemaa:
```suggestion
func TestReconcileInFlightTask_CloneTaskFailed(t *testing.T) {
```

internal/service/vmservice/power.go
mcbenjemaa, line 41: I imagine we can improve this further by checking that there is no TaskRef, we requeue until there is no TaskRef so we can proceed with `startVirtualMachine()`
this ensures that there is no ongoing tasks, although ReconcileInFlightTask does it, but this add more idempotency and protection to the reconciliation, we never know how concurrent reconciliation can affect that.

internal/service/vmservice/vm.go
```diff
@@ -115,7 +115,8 @@ func ReconcileVM(ctx context.Context, scope *scope.MachineScope) (infrav1.Virtua
 	} // VirtualMachineProvisioned reason is WaitingForBootstrapReady
 
 	// handle invalid state of the machine
-	if conditions.GetReason(scope.ProxmoxMachine, infrav1.ProxmoxMachineVirtualMachineProvisionedCondition) == infrav1.ProxmoxMachineVirtualMachineProvisionedVMProvisionFailedReason {
+	if conditions.GetReason(scope.ProxmoxMachine, infrav1.ProxmoxMachineVirtualMachineProvisionedCondition) == infrav1.ProxmoxMachineVirtualMachineProvisionedVMProvisionFailedReason ||
```
mcbenjemaa: can we make a helper function that does the check, something like `ProxmoxMachineHasVMProvisionFailedReason()`

internal/service/vmservice/vm_test.go
```diff
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"testing"
 
+	lutherproxmox "github.com/luthermonson/go-proxmox"
```
mcbenjemaa, line 24: I would follow the same import pattern, either `goproxmox`
or change the whole imports to this one.

internal/service/vmservice/vm_test.go
```diff
@@ -695,3 +696,187 @@ func TestReconcileVM_CloudInitRunning(t *testing.T) {
...
+	machineScope.ProxmoxMachine.Spec.AllowedNodes = []string{"node1", "node2"}
+	machineScope.ProxmoxMachine.Spec.NumSockets = ptr.To[int32](1)
+	machineScope.ProxmoxMachine.Spec.NumCores = ptr.To[int32](1)
+	machineScope.ProxmoxMachine.Spec.MemoryMiB = ptr.To[int32](1024)
+	machineScope.ProxmoxMachine.Spec.Disks = &infrav1.Storage{
+		BootVolume: &infrav1.DiskSize{
+			Disk:   "scsi0",
+			SizeGB: 50,
+		},
+	}
```
mcbenjemaa, line 727: can we move this block to a helper function?

internal/service/vmservice/vm_test.go
```diff
@@ -695,3 +696,187 @@ func TestReconcileVM_CloudInitRunning(t *testing.T) {
 	require.NoError(t, err)
 	require.Equal(t, infrav1.VirtualMachineStatePending, result.State)
 }
+
+// This test is supposed to test the entire state machine transition.
+func TestReconcileVM_EndToEnd(t *testing.T) {
```
mcbenjemaa:
```suggestion
func TestReconcileVM_ProxmoxMachineConditionsRounds(t *testing.T) {
```

or if you have a better name 👍🏻

mcbenjemaa (commented):
Overall looks good, just some remarks,
The test added in `vm_test` is great.

internal/service/vmservice/vm_test.go
wikkyk: or just _StateMachine

internal/service/vmservice/vm_test.go
65278: I would not do it because it has 0 reusability.

internal/service/vmservice/vm_test.go
65278: Problem is that both goproxmox and proxmox are already taken.

wikkyk: Perhaps we could just move the pkg/proxmox/* subpackages up a level.

internal/service/vmservice/vm_test.go
wikkyk: I agree with @65278, if you need to do it a second time then a helper is in order but just for one? naaah

wikkyk:
>even though we're only testing invariants

You never know, you know. It's not a bad thing to assert invariants.

internal/service/vmservice/power.go
wikkyk: @copilot turn this suggestion into a task issue for future implementation

wikkyk: #727

internal/service/vmservice/vm_test.go
wikkyk: #728

wikkyk (approved):
lgtm, @mcbenjemaa?

mcbenjemaa (approved):

PR #734, opened 2026-04-26
fix(proxmoxmachine): tolerate owner Machine NotFound to unblock stuck deletes

raul-cq:
## What this fixes

The `ProxmoxMachine` reconciler currently deadlocks any `ProxmoxMachine` whose owner `Machine` has been removed before the InfraMachine was finalized. Reproduction: take any cluster, remove a `Machine` out-of-band (manual finalizer strip, `cluster.x-k8s.io/force-delete-machine` annotation, or any other actor) before its `ProxmoxMachine` finishes finalizing. The `ProxmoxMachine` then sits in `Terminating` forever and blocks the parent `ProxmoxCluster` and `Cluster` from finalizing.

## Root cause

[`internal/controller/proxmoxmachine_controller.go#L97-L105`](https://github.com/ionos-cloud/cluster-api-provider-proxmox/blob/main/internal/controller/proxmoxmachine_controller.go#L97-L105):

```go
machine, err := util.GetOwnerMachine(ctx, r.Client, proxmoxMachine.ObjectMeta)
if err != nil {
    return ctrl.Result{}, err
}
```

`util.GetOwnerMachine` returns `nil, nil` only when there is **no** Machine OwnerRef at all. When the OwnerRef is present but the Machine object is gone, it calls `client.Get` and returns `apierrors.NotFound`. The reconciler returns that error verbatim, controller-runtime backs off exponentially, and the delete branch (`if !proxmoxMachine.ObjectMeta.DeletionTimestamp.IsZero()`) is never reached. The `MachineFinalizer` never comes off.

The log signature observed in production:
```
"Reconciler error" err="Machine.cluster.x-k8s.io \"<name>\" not found"
```
repeating every backoff cycle for hours.

This is a known class of issue: [`docs/Troubleshooting.md`](https://github.com/ionos-cloud/cluster-api-provider-proxmox/blob/main/docs/Troubleshooting.md) under "Machine deletion deadlock" already documents the manual workaround (strip the `proxmoxmachine` finalizer by hand). This PR addresses the underlying controller behaviour so the manual workaround is no longer needed.

## What this PR does

In `Reconcile`, when `GetOwnerMachine` returns `IsNotFound`:

- **If the `ProxmoxMachine` has a `DeletionTimestamp`:** remove the `MachineFinalizer` and return cleanly. A log line tells the operator to verify the underlying Proxmox VM, since a full `machineScope` cannot be constructed without the owner Machine and `vmservice.DeleteVM` requires the scope. This is the conservative choice: unblock the controller-level deadlock without taking destructive action against the hypervisor under partial state.
- **If the `ProxmoxMachine` is not being deleted:** log and return `nil` instead of error-looping. Either the OwnerRef will be corrected (which triggers another reconcile via the watch) or the `ProxmoxMachine` will be deleted (which triggers another reconcile via the delete event).

All other failure modes still propagate as before.

## Tests

Adds an envtest in `proxmoxmachine_controller_test.go` covering the deletion-with-missing-owner case: the test creates a `ProxmoxMachine` with a Machine OwnerRef pointing to a non-existent Machine and the `MachineFinalizer` set, deletes it (the API server records a `deletionTimestamp` since the finalizer is held), runs `Reconcile`, and asserts the object is reaped (i.e. the finalizer was removed).

```
$ make test WHAT=./internal/controller/...
ok  github.com/ionos-cloud/cluster-api-provider-proxmox/internal/controller  30.534s  coverage: 50.6% of statements
```

## Compatibility

- No CRD changes.
- No public API changes.
- The IONOS Troubleshooting doc workaround still works as a runtime escape hatch for clusters that were stuck before this fix is deployed.

## Why I'm sending this

We hit this in production (CAPMOX v0.8.1) when a workload cluster's apiserver became unreachable mid-delete and a Machine ended up reaped before its InfraMachine finalized — the exact deadlock above. Walking the diagnosis backward led here. Happy to iterate on the patch shape, the log wording, or the test scaffolding.

### DIFF
--- a/internal/controller/proxmoxmachine_controller.go
+++ b/internal/controller/proxmoxmachine_controller.go
@@ -97,6 +97,35 @@ func (r *ProxmoxMachineReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 	// Fetch the Machine.
 	machine, err := util.GetOwnerMachine(ctx, r.Client, proxmoxMachine.ObjectMeta)
 	if err != nil {
+		// The owner Machine OwnerRef is set but the Machine object itself is
+		// gone. This is the stuck-finalizer deadlock: without this branch the
+		// reconciler returns the NotFound error to controller-runtime, which
+		// requeues with exponential backoff forever, never reaching
+		// reconcileDelete and never removing our finalizer. The ProxmoxMachine
+		// then blocks the parent ProxmoxCluster (waiting for machines) and the
+		// Cluster from finalizing.
+		//
+		// We can hit this whenever the owner Machine is removed before the
+		// InfraMachine is finalized: a manual finalizer strip, the
+		// `cluster.x-k8s.io/force-delete-machine` annotation, or any other
+		// actor that out-of-band reaps the Machine.
+		if apierrors.IsNotFound(err) {
+			if !proxmoxMachine.ObjectMeta.DeletionTimestamp.IsZero() {
+				logger.Info("Owner Machine not found while ProxmoxMachine is being deleted; removing finalizer to unblock cluster delete. The underlying Proxmox VM may still exist and should be verified by an administrator.")
+				if ctrlutil.RemoveFinalizer(proxmoxMachine, infrav1.MachineFinalizer) {
+					if err := r.Client.Update(ctx, proxmoxMachine); err != nil {
+						return ctrl.Result{}, errors.Wrap(err, "removing finalizer after owner Machine not found")
+					}
+				}
+				return ctrl.Result{}, nil
+			}
+			// ProxmoxMachine is not being deleted. The OwnerRef points to a
+			// missing Machine — log and skip rather than error-loop. Either
+			// the OwnerRef will be corrected, or the ProxmoxMachine will be
+			// deleted, both of which trigger another reconcile.
+			logger.Info("Owner Machine not found and ProxmoxMachine is not being deleted; skipping reconcile.")
+			return ctrl.Result{}, nil
+		}
 		return ctrl.Result{}, err
 	}
 	if machine == nil {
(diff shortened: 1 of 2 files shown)


mcbenjemaa (commented):
I wonder how did you reproduce this?
as the InfraMachines will be deleted as long as the parent Machine is being deleted which is why it has an ownerReference.
in fact, the CAPI machine controller waits for the infraMachine to be deleted first, then cleanup the finalizer and deletes the Machine.

Do you have something specific that lets the parent machine gone, and the inframachine remain?

raul-cq:
Hi there... this seems to be a corner case. When creating a cluster one worker vm creation got stuck in cloud-init, i just destroyed the vm in proxmox, and then proceeded to delete the cluster. This is how i hit this behaviour.

internal/controller/proxmoxmachine_controller.go
```diff
@@ -97,6 +97,35 @@ func (r *ProxmoxMachineReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+		// reconciler returns the NotFound error to controller-runtime, which
+		// requeues with exponential backoff forever, never reaching
+		// reconcileDelete and never removing our finalizer. The ProxmoxMachine
+		// then blocks the parent ProxmoxCluster (waiting for machines) and the
+		// Cluster from finalizing.
+		//
+		// We can hit this whenever the owner Machine is removed before the
+		// InfraMachine is finalized: a manual finalizer strip, the
+		// `cluster.x-k8s.io/force-delete-machine` annotation, or any other
+		// actor that out-of-band reaps the Machine.
```
wikkyk, line 111: This comment (that looks AI-generated btw, please attribute it) is excessive. Convert the comments in your PR to one-liners and instead elaborate in the commit message.

65278:
Let me weigh in. Technically I like the idea, qmstop should reschedule because it can technically fail or time out. The problem is that we're depending on proxmox implementation defined behaviour to find the condition. I'll try to cook up a better usage of the api/UPID before I review this properly.

wikkyk:
We're looking into adding support for structured errors to go-proxmox so that we can do this on the capmox side properly.

PR #744, opened 2026-05-05
    Fix ProxmoxMachine deletion to wait for stop/destroy tasks

### LINKED ISSUE #742: v0.8.1: ProxmoxMachine deletion does not track async stop/destroy task failures; stale VM can be re-adopted after VM is running - destroy failed

While replacing a `MachineDeployment` worker in a Talos-based Cluster API workload cluster on Proxmox, I deleted a worker `Machine` whose backing VM/Talos identity had become stale.

The affected worker `Machine` was `qak8scapicl1-worker-g4f5d-xwrpb`. It had adopted/correlated to a stale control-plane Kubernetes `Node` instead of a worker node:

- CAPI `Machine`: `qak8scapicl1-worker-g4f5d-xwrpb` (worker Machine)
- Workload `NodeRef`: `qak8scapicl1-cp-640dc8f1-9btgj`
- Node IP: `192.168.160.2`
- The workload `Node` with the same provider identity had the Kubernetes `control-plane` role
- Direct Talos check on `192.168.160.2` showed `etcd` unhealthy (`Waiting` / `Fail`), and that node was not one of the three live etcd members

Deleting that Machine alone did not remove the underlying Proxmox VM. capmox/Proxmox logged the Proxmox destroy task failure:

```text
VM 103 is running - destroy failed
```

After that, a replacement worker `Machine` (`qak8scapicl1-worker-g4f5d-22dvn`) was created and re-used / re-adopted the same stale identity. This showed the problem was not just a stale Kubernetes `Node` object: the old backing Proxmox VM / Talos disk identity was still present and could be correlated to the replacement Machine.

The operator-visible result was dangerous: aggregate CAPI / `MachineDeployment` status could look green enough while a worker Machine was actually attached to a stale control-plane-role Kubernetes Node and stale Talos identity.

The manual workaround was:

1. Scale the affected `MachineDeployment` down to stop immediate replacement/adoption churn.
2. Manually stop/destroy the stale Proxmox VM whose destroy task had failed.
3. Remove any stale Kubernetes Node if it remained after the VM was gone.
4. Scale the `MachineDeployment` back up.
5. Verify the new worker Machine's `NodeRef`, providerID, hostname, role, InternalIP, and Talos identity all matched.

After manually destroying the stale VM and scaling back up, a new worker `Machine` came up cleanly:

- Machine / NodeRef: `qak8scapicl1-worker-g4f5d-59sln`
- providerID: `proxmox://7db57ef1-8c01-42bb-b1d2-0c9e600708c7`
- InternalIP: `192.168.160.2`
- Talos member hostname: `qak8scapicl1-worker-g4f5d-59sln`
- Talos machine type: `worker`

Final verification was healthy: exactly 3 control-plane Nodes and 3 worker Nodes, all Ready; exactly three healthy etcd members on the real control-plane nodes.

If Proxmox VM destruction fails, capmox should treat `ProxmoxMachine` deletion as incomplete until the VM is confirmed absent.

In particular:

- A failed Proxmox `qmdestroy` task should not be treated as successful VM cleanup.
- The provider should not allow the old IP, provider identity, or backing VM/Talos identity to be re-used or re-adopted by a replacement Machine until VM destruction is confirmed.
- The `ProxmoxMachine` condition should include a clear deletion failure reason and the Proxmox task exit status, for example: `qmdestroy: VM 103 is running - destroy failed`.
- Operators should see an obviously blocked deletion / cleanup failure instead of a replacement Machine silently adopting stale Talos/Kubernetes identity.
- Ideally, capmox should only remove deletion/finalizer state and release provider-owned identity/IP lifecycle after it has confirmed that the Proxmox VM is gone.

This looks like an async task handling gap in the deletion path.

I checked `v0.8.1` and current `main` (`ee96ba56f84fe47d400025c26b971d050702ced0`); the relevant deletion-path files appeared unchanged between the tag and current main at the time of checking.

Source-code observations:

1. `reconcileDelete()` calls `vmservice.DeleteVM()` and, if there is no immediate error, requeues with the comment `VM is being deleted`:

   https://github.com/ionos-cloud/cluster-api-provider-proxmox/blob/v0.8.1/internal/controller/proxmoxmachine_controller.go#L161-L174

   ```go
   err := vmservice.DeleteVM(ctx, machineScope)
   if err != nil {
       return reconcile.Result{}, err
   }
   // VM is being deleted
   return reconcile.Result{RequeueAfter: infrav1.DefaultReconcilerRequeue}, nil
   ```

2. `vmservice.DeleteVM()` removes the finalizer only when `DeleteVM()` returns an immediate `VMNotFound` / `ErrVMIDFree`. When the API call returns a Proxmox task successfully, it does not appear to persist that task UPID or inspect the eventual task result:

   https://github.com/ionos-cloud/cluster-api-provider-proxmox/blob/v0.8.1/internal/service/vmservice/delete.go#L34-L55

3. The Proxmox API client starts a stop task for a running VM, ignores that returned task, and immediately starts a delete/destroy task:

   https://github.com/ionos-cloud/cluster-api-provider-proxmox/blob/v0.8.1/pkg/proxmox/goproxmox/api_client.go#L215-L255

   ```go
   if vm.IsRunning() {
       if _, err = vm.Stop(ctx); err != nil {
           return nil, fmt.Errorf("cannot stop vm id %d: %w", vmID, err)
       }
   }

   task, err := vm.Delete(ctx)
   if err != nil {
       return nil, fmt.Errorf("cannot delete vm with id %d: %w", vmID, err)
   }

   return task, nil
   ```

   This seems racy: `vm.Stop(ctx)` submits an async Proxmox task, but capmox does not wait for it to finish before submitting `vm.Delete(ctx)`. Proxmox can accept the destroy task and later fail it with `VM 103 is running - destroy failed`.

4. capmox already has a task status helper for normal reconciliation. `taskservice.ReconcileInFlightTask()` checks `Status.TaskRef` and marks non-`qmstart` failed tasks with `TaskFailedReason` and the task exit status:

   https://github.com/ionos-cloud/cluster-api-provider-proxmox/blob/v0.8.1/internal/service/taskservice/task.go#L66-L147

   But the delete path does not store the stop/destroy task in `ProxmoxMachine.status.taskRef`, and `ReconcileVM()` is bypassed during deletion, so that existing failed-task handling is not applied to deletion tasks.

5. IPAM claims are owner-referenced to the `ProxmoxMachine`:

   https://github.com/ionos-cloud/cluster-api-provider-proxmox/blob/v0.8.1/pkg/kubernetes/ipam/ipam.go#L431-L510

   If the provider cannot guarantee VM absence before replacement, a new Machine can receive the same IP while the old VM/Talos identity is still alive.

Possible provider-side hardening:

- Make deletion task-aware, similar to normal VM reconciliation:
  - persist the stop task UPID in `status.taskRef` or a deletion-specific task reference;
  - requeue while the stop task is running;
  - only submit `qmdestroy` after the VM is confirmed stopped.
- Persist and monitor the destroy task UPID:
  - requeue while `qmdestroy` is running;
  - on success, confirm `CheckID` is free / VM lookup returns not found;
  - only then remove finalizer / cleanup state.
- If a stop or destroy task fails:
  - keep the finalizer;
  - set a deletion-specific failure condition with `task.Type` and `task.ExitStatus`;
  - do not silently continue replacement convergence until the VM is gone.
- Consider an explicit deletion state machine, for example: `Stopping` -> `WaitingForStopped` -> `Destroying` -> `WaitingForDestroyed` -> `Destroyed` / `DeletionFailed`.
- Add tests for deletion races:
  - running VM: `Stop()` returns task A, `Delete()` should not be called until task A succeeds;
  - destroy task accepted but later fails with `VM is running - destroy failed`: finalizer remains and condition shows deletion failure;
  - destroy task succeeds but `CheckID` / VM lookup still shows VM present: finalizer remains and reconcile continues;
  - destroy task succeeds and VMID is free / VM not found: finalizer is removed.

**Environment:**

- Proxmox version: 9.1.1
- Cluster-api-provider-proxmox version: `ghcr.io/ionos-cloud/cluster-api-provider-proxmox:v0.8.1`
- Kubernetes version: `v1.35.3` i
- OS (e.g. from `/etc/os-release`): Talos Linux v 1.12.6 workload nodes; 
- Cluster API version: `v1.12.4`
- Talos control-plane provider version: `v0.5.12`
- In-cluster IPAM provider version: `v1.0.3`
- IPAM: Cluster API in-cluster IPAM (`IPAddressClaim` / `IPAddress`)

Leonorus:
Issue #, if available:
    Fixes #742
    
    Description of changes:
    - Make ProxmoxMachine deletion task-aware for async Proxmox qmstop and qmdestroy tasks.
    - Return and track the qmstop task for running VMs instead of immediately submitting qmdestroy while the VM may still be running.
    - Store deletion task UPIDs in status.taskRef and requeue while stop/destroy tasks are still running.
    - Keep the finalizer and node location until deletion is confirmed by a successful destroy task and VMID-free check.
    - Surface deletion task failures with VirtualMachineProvisioned=False, Reason=DeletionFailed, and the Proxmox task exit status, e.g. qmdestroy: VM 103 is running - destroy failed.
    - Preserve safe cleanup paths for already-free VMIDs, missing/expired tasks, and unassigned/reserved VMIDs.
    - Add regression coverage for deletion task tracking, failed destroy handling, VMID-free confirmation, transient task lookup failures, and running-VM stop-before-destroy behavior.

### DIFF
--- a/.github/workflows/govulncheck.yml
+++ b/.github/workflows/govulncheck.yml
@@ -96,6 +96,6 @@ jobs:
         run: |
           jq '.runs[].tool.driver.rules |= map(if .properties != null and .properties.tags != null then .properties.tags |= unique else . end)' \
             govulncheck.sarif >govulncheck-fixed.sarif
-      - uses: github/codeql-action/upload-sarif@8aad20d150bbac5944a9f9d289da16a4b0d87c1e # v4.36.2
+      - uses: github/codeql-action/upload-sarif@54f647b7e1bb85c95cddabcd46b0c578ec92bc1a # v4.36.3
         with:
           sarif_file: govulncheck-fixed.sarif


internal/service/taskservice/task.go
```diff
@@ -55,14 +61,29 @@ func GetTask(ctx context.Context, machineScope *scope.MachineScope) (*proxmox.Ta
...
 
-	task, err := machineScope.InfraCluster.ProxmoxClient.GetTask(ctx, *machineScope.ProxmoxMachine.Status.TaskRef)
+	taskRef := *machineScope.ProxmoxMachine.Status.TaskRef
+	task, err := machineScope.InfraCluster.ProxmoxClient.GetTask(ctx, taskRef)
 	if err != nil {
-		return nil, ErrTaskNotFound
+		if isTaskNotFoundError(err) {
+			return nil, fmt.Errorf("%w: %w", ErrTaskNotFound, err)
+		}
+		return nil, fmt.Errorf("get task %s: %w", taskRef, err)
```
wikkyk:
```suggestion
		return nil, fmt.Errorf("received unknown task %s: %w", taskRef, err)
```

internal/service/taskservice/task.go
```diff
@@ -44,6 +45,11 @@ const (
 	TaskInfoStateError      = TaskInfoState("error")
 )
 
+// task type identifiers reported by Proxmox.
+const (
+	taskTypeQMStart = "qmstart"
+)
+
```
wikkyk: I don't think this is really needed, it only comes up once.

internal/service/vmservice/delete.go
```diff
@@ -18,43 +18,230 @@ package vmservice
...
+		storeDeletionTask(machineScope, task)
+		return true, nil
+	case task.IsSuccessful && task.IsCompleted:
+		clearTaskState(machineScope)
+		if task.Type == deletionTaskQMDestroy {
+			verificationContext := fmt.Sprintf("%s task %s succeeded", task.Type, task.UPID)
+			if completed, err := completeIfVMIDFree(ctx, machineScope, verificationContext); err != nil || completed {
+				return completed, err
+			}
+			setDeletionFailedCondition(machineScope, fmt.Sprintf("%s task succeeded but VMID %d is still in use", task.Type, machineScope.ProxmoxMachine.GetVirtualMachineID()))
```
wikkyk: This could happen when you schedule a deletion and a creation. The old VM gets deleted and the new VM gets created before this check runs and then you see that the deletion task succeeded and VMID is in use, it's just not _still_ in use, it's _again_ in use.

wikkyk:
Don't worry about the test 'failure' - it's only the sonarqube report step that's failing (I'm working on improving the workflow).

wikkyk:
Thanks. I'll let @65278 do another review because he knows that part of the code better.

wikkyk:
It's unfortunate that it's still done via a heuristic but that's the API that we're working with.

wikkyk:
We're trying to make this slightly better so I'll leave it open until then.

wikkyk:
We're looking into adding support for structured errors to go-proxmox so that we can do this on the capmox side properly.

Issue #724, opened 2026-04-14
Scheduler causes uneven CPU overcommit on heterogeneous Proxmox clusters


emanuelebosetti:
## Problem

On Proxmox clusters where nodes have different CPU core counts, the current round-robin scheduler distributes VMs evenly by count, ignoring CPU capacity. This causes severe CPU overcommit on nodes with fewer cores while nodes with more cores remain underutilized.

The scheduler only considers memory (`schedulerHints.memoryAdjustment`). CPU is never evaluated.

### Real-world example

A 20-node production Proxmox cluster hosting multiple CAPI-managed Kubernetes clusters (229 VMs total across different clusters) with 3 node classes:

| Class | Nodes | Cores | RAM |
|---|---|---|---|
| A | 10 | 64 | 504 GB |
| B | 2 | 32 | 661 GB |
| C | 8 | 32 | 504 GB |

Current state:

| Node class | Avg CPU ratio | Max CPU ratio |
|---|---|---|
| A (64 cores) | 2.67:1 | 3.25:1 |
| B (32 cores, 661 GB RAM) | 2.28:1 | 2.81:1 |
| C (32 cores, 504 GB RAM) | 1.41:1 | 1.97:1 |

After a rollout restart (all 229 VMs re-scheduled from zero):

| Metric | Result |
|---|---|
| CPU ratio max | **4.44:1** |
| CPU ratio min | 1.38:1 |
| Spread (max-min) | **3.06** |
| Nodes above 3:1 | **9 / 20** |

Class B nodes (32 cores, highest RAM) are hit the hardest: the scheduler prefers them because they have the most available memory, but they have half the CPU cores. The imbalance persists until manual rebalancing via Proxmox live migration.

### Root cause

In `vmscheduler.go`, `selectNode`:
1. Sorts nodes by available memory (descending)
2. Counts existing VMs per node
3. Picks the node with the fewest VMs that has enough memory

CPU capacity (`node.CPUInfo.CPUs`) and CPU allocation per VM (`vm.CPUs`) are available from the Proxmox API but never queried by the scheduler.

### Suggestion

Introduce a `schedulerHints.cpuAdjustment` field (analogous to `memoryAdjustment`) that, when set, enables a scheduling mode that considers both CPU and memory saturation when placing VMs. When disabled (default `0`), the current round-robin behavior is preserved.

[...]

PR #747, opened 2026-05-08
feat(scheduler): tolerance-weighted CPU/memory aware VM scheduling

emanuelebosetti:
*Issue #, if available:* #724

*Description of changes:*

Map memory-first priority into a tolerance-weighted CPU/memory scheduler so heterogeneous Proxmox clusters stop concentrating VMs on smaller-CPU nodes by accident.

**How it works**

When `schedulerHints.cpuAdjustment` is set to a non-zero value, the scheduler switches from round-robin to a tolerance-weighted scoring that considers both CPU and memory headroom on every candidate node:

- Per-resource score is shaped by a non-linear curve (`saturationExponent = 2.0`) so already-saturated hosts are heavily penalized.
- Free capacity in the physical range counts fully (1 point/unit); free capacity in the overcommit range is discounted (`overbookDiscount = 0.5`).
- Per-resource weights are derived from `memoryTolerance` and `cpuTolerance` via `weight = (100 − tolerance) / 100`. High tolerance = operator OK with that resource saturating = less influence on the decision.
- CPU and memory remain hard-fit constraints; nodes that cannot fit the request are excluded upfront with `InsufficientCPUError` / `InsufficientMemoryError`.
- Anti-clumping: existing placements recorded in `status.NodeLocations` inflate a node's effective usage, so VMs cloned in rapid succession (Proxmox API takes seconds to reflect just-cloned VMs) don't all converge on the same target.

**API changes (additive, backwards-compatible)**

| Field | Type | Default | Meaning |
|---|---|---|---|
| `cpuAdjustment` (existing) | `*int64` | `0` | Overcommit factor; non-zero enables CPU-aware scheduling |
| `memoryTolerance` (new) | `*int64` (0..100) | `0` | Lower = more protective of memory |
| `cpuTolerance` (new) | `*int64` (0..100) | `100` | Higher = more acceptable to saturate CPU |

`cpuAdjustment: 0` (default) reproduces today's memory-only round-robin — full backward compatibility.

**A note on the chosen defaults**

The defaults ship aligned with the memory-first priority discussed in the issue thread for HA workloads: when CPU-aware scheduling is enabled, memory is fully protected and CPU acts only as a hard-fit constraint until the operator opts in to spreading.

These values *deliberately differ* from those I proposed earlier in the thread (`memoryTolerance: 100`, `cpuTolerance: 0`). Re-checking that proposal end-to-end, those values are inverted from the priority discussed earlier and would have caused the default CPU-aware mode to ignore memory headroom. Happy to revisit if you'd rather keep my original proposed defaults — the change is two constants and a couple of test fixtures.

*Testing performed:*

- Unit tests for tolerance scoring covering: defaults follow memory, opposite polarity follows CPU, overbook penalty, hard-fit constraints (CPU and memory), heterogeneous-cluster regression, anti-clumping under rapid scale.
- `make lint test verify` pass locally (envtest 1.33.0). The `pkg/convert/sentinel.go` `goconst`/`nolintlint` warnings are pre-existing on `main` and unrelated to this PR.
- Algorithm coverage on `internal/service/scheduler` is **87.1%**.
- **Validated on a real Proxmox cluster** (10 nodes across two physical hosts): triggered `MachineDeployment` rollouts on a 5-machine workload cluster with `cpuAdjustment: 200` and confirmed `selectNodeByToleranceScoring` runs on every new VM, applies the documented defaults (`memoryTolerance=0 cpuTolerance=100` → `cpuWeight=0`, memory drives), and the anti-clumping `NodeLocations` inflation correctly reduces a node's score after each placement (observed `scoreMem` drop from `0.807` to `0.762` between two consecutive decisions on the same node within ~1s).

**Attribution**

Developed with assistance from Claude (Opus 4.7), as per the AI attribution practice followed in previous PRs.


api/v1alpha1/proxmoxcluster_types.go
wikkyk, line 1: Please don't change v1alpha1 because it's frozen.

api/v1alpha2/proxmoxcluster_types.go
```diff
@@ -185,6 +185,39 @@ type SchedulerHints struct {
...
+	// When set to a non-zero value, the scheduler switches from round-robin placement
+	// to a tolerance-weighted scoring that considers both CPU and memory headroom.
+	// See memoryTolerance and cpuTolerance for preference tuning.
+	// +kubebuilder:validation:Minimum=0
+	// +optional
+	CPUAdjustment *int64 `json:"cpuAdjustment,omitempty"`
+
+	// memoryTolerance expresses how acceptable it is to saturate memory on a node,
+	// on a scale from 0 (protect memory at all costs) to 100 (memory saturation is fine).
+	// Internally it is inverted into a weight: weight = (100 - memoryTolerance) / 100.
```
wikkyk: Here and under CPUTolerance I would omit the internal calculation. Implementation details may change and they're irrelevant to the user.

wikkyk: also this whole comment and the CPUTolerance comment are a bit too verbose and llm-like. there is no need to mention the scoring etc

internal/service/scheduler/vmscheduler.go
```diff
@@ -72,28 +95,66 @@ func selectNode(
...
+			Name:              nodeName,
+			AvailableMemory:   mem,
+			AllocatableMemory: allocMem,
+			AvailableCPUs:     cpus,
+			AllocatableCPUs:   allocCPUs,
+		}
 	}
 
+	requestedMemory := uint64(ptr.Deref(machine.Spec.MemoryMiB, 0)) * 1024 * 1024
+	requestedCPUs := int(ptr.Deref(machine.Spec.NumSockets, 0)) * int(ptr.Deref(machine.Spec.NumCores, 0))
```
wikkyk: I wouldn't cast to `int` here, just use the dereferenced types of Num{Socket,Cores}.

internal/service/scheduler/vmscheduler.go
```diff
@@ -72,28 +95,66 @@ func selectNode(
...
+			ctx, nodes, locations, requestedMemory, requestedCPUs,
+			memoryAdjustment, cpuAdjustment,
+			schedulerHints.GetMemoryTolerance(), schedulerHints.GetCPUTolerance(),
+		)
+	}
+
+	return selectNodeByRoundRobin(ctx, nodes, locations, requestedMemory)
+}
+
+// selectNodeByRoundRobin is the legacy scheduling algorithm: round-robin by VM count with memory as the only constraint.
```
wikkyk: I don't really like having 2 scheduling algos. Plus, most of the change just removes comments.

internal/service/scheduler/vmscheduler.go
```diff
@@ -31,6 +32,15 @@ import (
...
 
+const (
+	// overbookDiscount is the per-unit score of headroom in the overcommit range,
+	// relative to 1 point per unit in the physical range.
+	overbookDiscount = 0.5
+	// saturationExponent shapes the per-resource score as a power curve, so that
+	// nodes closer to saturation score disproportionately less than empty nodes.
+	saturationExponent = 2.0
+)
+
```
wikkyk, line 43: How did you arrive at these constants? They look pretty aggressive. I know that I said I wanted to punish overcommit when physical resources are available but maybe not that hard :)

wikkyk:
I like where this is going but it needs some work.

Please let's only have one scheduling algo. 
You can still use the new algo when CPU-aware placement is disabled. Overcommit handling being on a curve rather than linear is useful. For small overcommit it's hardly a visible change and for high overcommit is an improvement. You shouldn't need to enable CPU-aware placement to get the benefit.
Plus, two algos are more expensive to maintain than one.
Behaviour is more difficult to reason about when there are two behaviours to consider, which is extra brain load for ops.

api/v1alpha1/conversion.go
```diff
@@ -193,6 +193,14 @@ func Convert_v1alpha2_ProxmoxClusterSpec_To_v1alpha1_ProxmoxClusterSpec(in *v1al
 	return nil
 }
 
+// Convert_v1alpha2_SchedulerHints_To_v1alpha1_SchedulerHints handles the lossy
+// conversion of SchedulerHints from v1alpha2 to v1alpha1. MemoryTolerance and
+// CPUTolerance are v1alpha2-only fields that are dropped on down-conversion and
+// restored from annotation on up-conversion via restoreSchedulerHints.
```
wikkyk: You don't really need that comment.

api/v1alpha2/proxmoxcluster_types.go
```diff
@@ -185,6 +185,33 @@ type SchedulerHints struct {
 	// +kubebuilder:validation:Minimum=0
 	// +optional
 	MemoryAdjustment *int64 `json:"memoryAdjustment,omitempty"`
+
+	// cpuAdjustment allows to adjust a node's CPU capacity by a given percentage.
+	// For example, setting it to 300 allows to allocate 300% of a host's CPU cores for VMs (3:1 overcommit),
```
wikkyk:
```suggestion
	// For example, setting it to 300 allows allocating 300% of a host's CPU cores for VMs (3:1 overcommit),
```

This comment could also explain what happens when you set the field to less than 100 (e.g. 95 to follow MemoryAdjustment's example).

api/v1alpha2/proxmoxcluster_types.go
```diff
@@ -185,6 +185,33 @@ type SchedulerHints struct {
...
+	// and setting it to 100 limits CPU allocation to the physical core count (no overcommit).
+	// Setting it to 0 entirely disables scheduling CPU constraints (default).
+	// When set to a non-zero value, the scheduler switches from round-robin placement
+	// to a tolerance-weighted scoring that considers both CPU and memory headroom.
```
wikkyk: First of all the old algo wasn't quite round-robin (it was memory-aware) so documenting it as round-robin now is just adding to confusion.
Second of all, as I explained before, I would prefer to only have 1 scheduling algo.

wikkyk:
With the defaults of `MemoryTolerance=0` and `CPUTolerance=100` plus the curve-based overcommit scoring, we don't really need to switch tolerance on overcommit. Instead, we can just always use the curve-based algo, which would also simplify it as there is no need for the `cpuAware` branches then.

api/v1alpha1/proxmoxcluster_types.go
emanuelebosetti: Removed three new field in v1alpha1

api/v1alpha2/proxmoxcluster_types.go
emanuelebosetti: Omited internal calculation on MemoryTolelrance and CPUTollerance, now the description expose only information relevant to the user for cluster configuration

internal/service/scheduler/vmscheduler.go
emanuelebosetti: Done — requestedCPUs is now int32

internal/service/scheduler/vmscheduler.go
emanuelebosetti: These were picked against the heterogeneous cluster test (24 VMs across mixed 64/32-core nodes): legacy lands at 4.4x CPU spread, current constants bring it to 1.43x. What shape did you have in mind?

emanuelebosetti:
Unified into a single algorithm — `selectNodeByRoundRobin` is gone, `selectNode` always runs the curve-based scoring. With `cpuAdjustment=0` CPU is not fetched, not enforced, and contributes 0 to the score, so the algorithm naturally degrades to memory-only without a separate code path. Unit tests cover the heterogeneous case and the rapid-scale anti-clumping scenario; I'd like to validate it on a real Proxmox cluster tomorrow before considering this fully done.

internal/service/scheduler/vmscheduler.go
```diff
@@ -72,86 +113,207 @@ func selectNode(
 	allowedNodes []string,
 	schedulerHints *infrav1.SchedulerHints,
 ) (string, error) {
-	byMemory := make(sortByAvailableMemory, len(allowedNodes))
+	memoryAdjustment := schedulerHints.GetMemoryAdjustment()
+	cpuAdjustment := schedulerHints.GetCPUAdjustment()
+	cpuEnabled := cpuAdjustment > 0
```
wikkyk, line 118: I don't think cpu needs any special treatment. The algo should just do the right thing for both memory and cpu and work the same way for both resources.

wikkyk: In the ideal world, right after reading the counts and adjustments, the algo should be generic. This will make it easy to add more kinds of resources to SchedulerHints, e.g. storage.

internal/service/scheduler/vmscheduler.go
wikkyk: The way I read this algo, these consts produce an exponential decay curve (decreases aggressively early on and then tapers off). True, this does produce a very nice and evened-out allocation when the nodes are under-booked. However, as load increases, the contribution of any additional load has a weaker effect on the score.
I was thinking of something closer in shape to the first quarter of a cosine curve where the contribution of any additional load has a stronger effect on the score as load increases. This would mean a less even allocation when under-booked but any over-booked machines are very sharply scored down.
If anything, I'd flip these values.

internal/service/scheduler/vmscheduler.go
```diff
@@ -72,86 +113,207 @@ func selectNode(
...
 
+	if len(candidates) == 0 {
+		return "", buildInsufficientError(nodes, requestedMemory, requestedCPUs)
+	}
+
+	sort.SliceStable(candidates, func(i, j int) bool {
+		// primary: curve-based score (higher wins); tie-break: raw memory
+		// headroom (kicks in when many nodes saturate under pending inflation).
+		if candidates[i].scoreTotal != candidates[j].scoreTotal {
+			return candidates[i].scoreTotal > candidates[j].scoreTotal
```
wikkyk, line 214: You've inverted the comparison (`SliceStable` takes a 'less' function but you are returning the greater value). This needs to be documented better, it's easy to miss; although I would prefer for this algo to work properly given the correct sorting direction.

internal/service/scheduler/vmscheduler.go
emanuelebosetti: I extended the math analysis with Claude to also evaluate a logarithmic family, because once you've moved off `r²` there's a practical issue worth weighing before committing to the cosine shape. Sharing the full picture below.

**How `raw` is computed today**

In `resourceScore`, before applying any curve, we compute a single "free capacity ratio" per resource:

```go
raw := (freePhys + overbookDiscount*freeOver) / phys
```

with `freePhys = max(0, phys − used)` and `freeOver = max(0, alloc − max(used, phys))`. The split is what gives the physical zone its full point per unit and the overcommit zone only half a point per unit. A worked example for a 32-core node with `cpuAdjustment: 300` (so `alloc = 96`):

| `used` | `freePhys` | `freeOver` | `raw` |
|---:|---:|---:|---:|
| 0 (empty) | 32 | 64 | **2.0** |
| 16 | 16 | 64 | 1.5 |
| 32 (physical limit hit) | 0 | 64 | 1.0 |
| 64 (deep into overcommit) | 0 | 32 | 0.5 |
| 96 (full) | 0 | 0 | 0.0 |

**`raw` can exceed 1.** Without overcommit (`adjustment = 100`) we have `alloc = phys`, so `freeOver = 0` and `raw ∈ [0, 1]`. The moment overcommit is enabled (`adjustment > 100`) the empty-node `raw` is `1 + 0.5·(adjustment/100 − 1)`, which is 1.5 at 2:1, 2.0 at 3:1, 3.0 at 5:1. The current `r²` curve doesn't care since it's monotonically increasing for any real `r`, but it's a constraint to keep in mind when picking the replacement.

**Curve candidates side by side**
Raw 2.0 Empty Node,  Raw 0 Full Node

Image

| `raw` | `r²` (current) | `sin(r·π/2)` | `log₂(1+r)` | `log(1+9r)/log(10)` |
|---:|---:|---:|---:|---:|
| 0.0 | 0.000 | 0.000 | 0.000 | 0.000 |
| 0.1 | 0.010 | 0.156 | 0.138 | 0.279 |
| 0.3 | 0.090 | 0.454 | 0.379 | 0.568 |
| 0.5 | 0.250 | 0.707 | 0.585 | 0.740 |
| 0.7 | 0.490 | 0.891 | 0.766 | 0.863 |
| 0.9 | 0.810 | 0.988 | 0.926 | 0.959 |
| 1.0 | 1.000 | 1.000 | 1.000 | 1.000 |
| 1.5 | 2.250 | **0.707** | 1.322 | 1.161 |
| 2.0 | 4.000 | **0.000** | 1.585 | 1.279 |

The `sin` rows past `r=1` are the issue. The sine peaks at `r=1` and folds back, so a fully empty overcommit-enabled node ends up scoring *worse* than a half-loaded one. To use the sine we'd need to either clamp `raw` at 1 (which makes two empty nodes with different overcommit settings indistinguishable) or re-normalize the denominator so `raw` stays in `[0,1]` by construction. Both work, both are small changes, but they're extra cognitive load for anyone reading the algorithm.

The logarithm avoids the issue: `log₂(1+r)` and `log(1+9r)/log(10)` are monotonic on the whole real line — the plot just keeps climbing into the overcommit zone, no special casing needed. `k=9` controls how aggressive the curvature is.

**Slope behaviour (the "discrimination" signal)**

Slope at the two ends of the physical range tells you how aggressively the scheduler differentiates near-empty nodes vs. near-saturated nodes. Higher ratio = closer to the "sharply scored down when over-booked, less even when under-booked" property you described:

| Curve | slope at `r=0` | slope at `r=1` | ratio | shape |
|---|---:|---:|---:|---|
| `r²` | 0.00 | 2.00 | inverted | concave up — wrong direction |
| `sin(r·π/2)` | 1.57 | 0.00 | ∞ | concave down, flat plateau at `r=1` |
| `log₂(1+r)` | 1.44 | 0.72 | 2:1 | concave down, mild plateau |
| `log(1+9r)/log(10)` | 3.91 | 0.39 | 10:1 | concave down, sharper plateau |

The cosine quarter is the most extreme implementation of the property you described (literal "first quarter of a cosine"). The logarithm with `k=9` is close in spirit (10:1 ratio, plateau at the top) and removes the overcommit edge case entirely — at the cost of a less symmetric shape.

**Two viable directions**

1. **Cosine quarter**, with the denominator re-normalized to keep `raw ∈ [0,1]`. Most faithful to your original description, requires a small touch-up in `resourceScore` to handle overcommit.
2. **Logarithm**, `log(1 + k·r) / log(1 + k)` with `k=9` as default. Monotonic for any `raw`, no edge cases, tunable steepness via a single constant — but the shape diverges from "cosine".

Which way would you go?

wikkyk: This analysis isn't useful at all. The chart contradicts the table. The actual behaviour seems to contradict the analysis. Can *you*, Emanuele Bosetti, explain how the algorithm that you submitted works?

emanuelebosetti: let's consider the algorithm variables:
 
freePhys = max(0, phys - used)     --> physical free resource (ram or cpu)      
freeOver = max(0, alloc - max(used, phys))  --> free resource in the overcommit zone (ram or cpu)
raw  = (freePhys + 0.5 * freeOver) / phys --> available resource index (overcommit resource disadvantaged)
 
the closer "raw" is to 0, the less i want to allocate VM on the node.
 
e.g. with 4 CPU on the node, with a ratio of 3:1 (up to 12 vCPU usable), in case i have 0% usage we have
freePhys = max(0, 4-0) = 4
freeOver = max (0, 12 - max(0, 4)) = max (0, 12-4) = 8
raw =  (4 + 0.5 * 8)/ 4 = 2
 
with the same logic, if i have the 4 physical CPU full, same ratio of overcommit, i have
freePhys = max(0, 4-4) = 0
freeOver = max (0, 12 - max(4, 4)) = max (0, 12-4) = 8
raw =  (0 + 0.5 * 8)/ 4 = 1
 
below this threshold I want to penalise the resulting score, since I'm in overcommit
  
if the node is completely full (both physical and overcommit) i have
freePhys = max(0, 4-4) = 0
freeOver = max (0, 12 - max(12, 4)) = max (0, 12-12) = 0
raw =  (0 + 0.5 * 0)/ 4 = 0

This logic is used to calculate each resource "raw" index. 
I introduced the additional "score" function in order to penalise overcommited resources when the two dimensions (ram and cpu) are combined. 
 
As you correctly said the exponential function gives a boost too high when "raw" is over 1 (overcommit point in ratio 3:1)
so i tried the shifted logarithmic function in order to have a function always defined in the space [0; +infinite], with a plateu on "high raw index values". the base idea is: if the node is not overcommited, i do not need a big differentiation (i'm in the plateau of the function). if the node is overcommited (raw index --> 0) i'm highly penalizing.

The code hasn't yet been modified for the logarithmic function and still uses the exponential function while waiting to define how to use the "score" function.

wikkyk: There is a bit of danger when the cluster is highly heterogeneous.

Consider the scenario where you have a cluster with two machines, one has double the resources and the function gives it double the score. The larger node will keep getting selected until its score is brought down to below the empty smaller node's.

If you care about diverse assignment, then perhaps normalising on [0,1] by calculating each node's available capacity as a percentage of its total capacity is the right thing to do.

You'd either need to loop through the list sorted by the normalised load, or pre-select only the nodes that have sufficient capacity to allocate a VM. This is because the node with the lowest relative load may not have sufficient resources to host a given VM, so a large VM would still need to be allocated to a larger node even if it's not the least-loaded node. Basically, assigning to the lowest-relatively-loaded node that actually can fit the VM.

emanuelebosetti: > Consider the scenario where you have a cluster with two machines, one has double the resources and the function gives it double the score. The larger node will keep getting selected until its score is brought down to below the empty smaller node's.

I'm afraid I'm missing something out.
The raw score is actually not normalized, but it's already a dimentionless ratio, since the value is divided by phys (raw = (freePhys + 0.5 * freeOver) / phys )
e.g empty node with 4 and 64 CPUs - with same overecommit ratio of 3:1 - will have same "raw" value
4 CPU
freePhys = max(0, 4-0) = 4
freeOver = max (0, 12 - max(0, 4)) = max (0, 12-4) = 8
raw = (4 + 0.5 * 8)/ 4 = 2

64 CPU
freePhys = max(0, 64-0) = 64
freeOver = max (0, 192 - max(0, 64)) = max (0, 192-64) = 128
raw = (64 + 0.5 * 128)/ 64 = 2

>You'd either need to loop through the list sorted by the normalised load, or pre-select only the nodes that have sufficient capacity to allocate a VM. This is because the node with the lowest relative load may not have sufficient resources to host a given VM, so a large VM would still need to be allocated to a larger node even if it's not the least-loaded node. Basically, assigning to the lowest-relatively-loaded node that actually can fit the VM.

Moreover, the score calculation is already performed simulating the placement of the new VM on the candidate nodes (vmscheduler.go:186. here the algorithm is still divided on ram and cpu, i will then merge the dimensions as per your previous comment).
This should prevent that the node chosen cannot then actually fit the VM. 
Considering the two nodes in previous example, if I want to place a VM of 16 CPU, i would have:

4 CPU Node - 16 Core VM Placement
freePhys = max(0, 4-16) = 0
freeOver = max (0, 12 - max(16, 4)) = max (0, 12-16) = 0
raw = (0 + 0.5 * 0)/ 4 = 0 --> value is 0, the vm does not fit candidate node, also if empty

64 CPU Node - 16 Core VM Placement
freePhys = max(0, 64-16) = 48
freeOver = max (0, 192 - max(16, 64)) = max (0, 192-64) = 128
raw = (48 + 0.5 * 128)/ 64 = 1,75 --> it fit candidate node


This would not change also if the 4CPU node is completely empty and the 64CPU node not: the initial score would be better for the empty nodes, but simulating with the new vm on it will lead to 4CPU node discard because does not fit.

emanuelebosetti: Hi @wikkyk, picking this back up. Before I take it further, which direction would you like me to go? Once you point me the way, I'll move on it.

Thanks!

PR #756, opened 2026-05-15
Remove direct github.com/pkg/errors usage

### LINKED ISSUE #654: Stop using github.com/pkg/errors as it is unmaintained
As per title.
It's a larger job that might seem, will need some care to replace it with something strictly better.

MukundaKatta:
## Description
- replace direct `github.com/pkg/errors` imports with the standard library `errors` and `fmt`
- preserve wrapping with `%w` where callers rely on `errors.Is` or `errors.As`
- keep nil-wrap behavior in requeue paths that previously depended on `errors.Wrap(nil, ...)` returning nil
- move `github.com/pkg/errors` from direct to indirect in `go.mod`, since it is still pulled in transitively

Fixes #654

## Testing
- `go test ./internal/service/vmservice ./internal/inject ./pkg/cloudinit ./pkg/ignition ./pkg/kubernetes/ipam ./pkg/proxmox/goproxmox ./pkg/scope`
- `go test ./...` (fails locally because envtest cannot find `/usr/local/kubebuilder/bin/etcd`; non-envtest packages run and pass)
- `git diff --check`

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -22,7 +22,6 @@ require (
 	github.com/luthermonson/go-proxmox v0.4.1
 	github.com/onsi/ginkgo/v2 v2.28.3
 	github.com/onsi/gomega v1.40.0
-	github.com/pkg/errors v0.9.1
 	github.com/spf13/cobra v1.10.2
 	github.com/spf13/pflag v1.0.10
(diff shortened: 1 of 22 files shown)


wikkyk:
Please follow the AI contribution rules.

PR #770, opened 2026-05-26
test: use GinkgoHelperGo in suite tests and harden enrich_test type assertions

wikkyk:
*Issue #, if available:*

N/A — follow-up to #768 (onsi group bump).

*Description of changes:*

- Bump `github.com/onsi/ginkgo/v2` from v2.28.3 → v2.29.0 and `github.com/onsi/gomega` from v1.40.0 → v1.41.0 on this branch (same versions already merged to `main` via #768).
- Replace the inline `go func() { defer GinkgoRecover(); ... }()` pattern in `internal/controller/suite_test.go` and `internal/webhook/webhook_suite_test.go` with a named `startTestManager()` helper that calls `GinkgoHelperGo()`. This causes Ginkgo to attribute manager startup failures to the `BeforeSuite` call site rather than inside the anonymous function, and removes the need for a manual `defer GinkgoRecover()`.
- In `pkg/ignition/enrich_test.go`, replace the single chained `.(map).(map).([]any)` type assertion with separate `require.True` / `require.IsType` steps. Each step now fails with a clear message instead of a cryptic panic if the JSON structure is unexpected.

*Testing performed:*

- `make test` (pending CI)


wikkyk:
Closing in favour of a clean branch without release/v0.9.0 changes.

PR #779, opened 2026-06-02
Uniq gateway hotfix

65278:
*Issue #, if available:*
No issue available. Failure in our production.

*Description of changes:*
v1alpha2 was supposed to allow multiple network interfaces with only one gateway, but the network rendering code blocked this. I've refactored it (mostly manually, Claude wrote the commit messages).

*Testing performed:*
Still testing.


pkg/cloudinit/network.go
```diff
@@ -197,103 +182,121 @@ func (r *NetworkConfig) validate() error {
...
+		var errPrefix error
+
+		switch ptr.Deref(route.To, "") {
+		case "":
+			// Route without a target makes no sense.
+			return ErrMalformedRoute
+		case "default":
+			*hasGateway = true
+			// TODO: I can only guess address family here.
+			prefix, _ = netip.ParsePrefix("0.0.0.0/0")
```
wikkyk: could probably just set prefix to 0 here

pkg/cloudinit/network.go
```diff
@@ -197,103 +182,121 @@ func (r *NetworkConfig) validate() error {
...
 			if err != nil {
 				return ErrMalformedRoute
 			}
 		}
+
+		// Todo: IPFamily might collide for default routes or routes without
+		// explicit ipfamily.
+		// A route is uniquely identified by its normalized subnet, metric and table.
+		serialized := fmt.Sprintf("%s %d %d", prefix.String(), ptr.Deref(route.Metric, 0), ptr.Deref(route.Table, 0))
+		routeID := sha256.Sum256([]byte(serialized))
```
wikkyk: I can't say I _like_ it but it's reasonable for a hotfix.

wikkyk (dismissed):

pkg/cloudinit/network.go
65278: This is actually an unsolved problem. If you set ipv6 gateway "default", then you get an ipv4 gateway. The code would need to be smarter, but I haven't figured how.

pkg/cloudinit/network.go
65278: It's much better than the old logic. Consider that this is more correct and more efficient.

wikkyk:
Merged #792 here.

pkg/ignition/network.go
```diff
@@ -191,19 +231,25 @@ func render(name string, tpl string, data types.NetworkConfigData) ([]byte, erro
...
+					break
 				}
 			}
 		}
+		// adjust VRF routes by adding the routing table to each member route.
+		// This is to keep approximate expected behaviour with netplan v2, but
+		// could use its own validation pass in the future.
+		for j := range data[i].Routes {
+			data[i].Routes[j].Table = data[i].Table
+		}
```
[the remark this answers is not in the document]
wikkyk: The behaviour is deliberate. The comment serves as a reminder to implement better validation. I'll open an issue.

wikkyk (approved):

PR #720, opened 2026-04-13
feat: implement CAPI failure domain contract for zones

emanuelebosetti:
Issue: #717  
**Description of changes:**                                                                                                                                                                                                                                                
                                                                                                                                                                                                                                                                             
  Map existing CAPMOX zones to CAPI failure domains so that KubeadmControlPlane automatically distributes control plane nodes across Proxmox zones.                                                                                                                          
                  
  **API changes (additive, backwards-compatible):**                                                                                                                                                                                                                          
  - Add `Nodes []string` to `ZoneConfigSpec` for per-zone node lists
  - Add `ControlPlane *bool` to `ZoneConfigSpec` (defaults to `true`)                                                                                                                                                                                                        
  - Add `FailureDomains` to `ProxmoxClusterStatus` (CAPI v1beta2 contract)
  - Add `spec.failureDomain` to `ProxmoxMachineSpec` (InfraMachine contract)                                                                                                                                                                                                 
  - Add `GetZoneNodes()` helper on `ProxmoxCluster`
  - Add `FailureDomainNotReady` condition reason                                                                                                                                                                                                                             
                  
  **Controller changes:**                                                                                                                                                                                                                                                    
  - Add `reconcileFailureDomains` in cluster controller (sorted by name to prevent reconcile loops)
  - Read `Machine.Spec.FailureDomain` in machine controller with retryable `FailureDomainNotReady` condition
  - Use `effectiveZone`/`effectiveAllowedNodes` on `MachineScope` to avoid spec mutation from the controller                                                                                                                                                                 
  - Pass zone override to `GetInClusterPools` for IPAM pool selection
  - Set `Zone` on `NodeLocation` in `createVM`                                                                                                                                                                                                                               
  - Update scheduler to use `GetEffectiveAllowedNodes()`
                                                                                                                                                                                                                                                                             
  **Conversion:** 
  - Restore `Status.FailureDomains` and `Spec.FailureDomain` in v1alpha1 conversion webhook
  - Add manual conversion stub for `ProxmoxMachineSpec.FailureDomain`                                                                                                                                                                                                        
   
  **Testing performed:**                                                                                                                                                                                                                                                     
  - Unit tests for API types, scheduler, VM service helpers, and machine scope
  - Validated on a real Proxmox cluster (10 nodes, 2 zones) with 3 CP + 2 worker nodes                                                                                                                                                                                       
  - KCP correctly distributes CP replicas round-robin across failure domains
[...]

PR #784, opened 2026-06-04
Port of previous PR: support VM templates on local storage per node

### LINKED ISSUE #451: Local storage templates per Proxmox Node
[A clear and concise description of what you want to happen.]
Having multiple clustered proxmox nodes which only have local storage.

Currently only 1 template is supported for instances, which makes possible to deploy kubernetes nodes only on one proxmox node. 

0.7.0 release implemented `templateSelector` feature, but this is not the expected feature, as it will not work if 2 templates are found (one on each node) 

If template is on shared storage, but for kubernetes i want to use local storage Clone operation fails with error about unable to clone to local storage (which is quite strange, as from proxmox UI you can do full clone to local storage)

### Proposal:
Use current `templateSelector`, but when searching for template return map[string]int as node-template association. And when vm is being cloned, add specific node template as source. 

**Environment:**

- Cluster-api-provider-proxmox version: 0.7.0
- Kubernetes version: (use `kubectl version`): any
- OS (e.g. from `/etc/os-release`):  talos linux

chrodriguez:
**Issue**

Fixes #451

**Description of changes:**

Port of PR #467 (by @Atoms) rebased onto current main and adapted to integrate with the `TemplateMatchPolicy` feature (exact/subset/best) that was added to main after that PR was opened. 

Adds support for VM templates stored on local (non-shared) Proxmox storage, enabling clusters where each node has its own copy of the template:

- New `localStorage` boolean field in `TemplateSource` (v1alpha1 and v1alpha2)
- `FindVMTemplatesByTags` replaces `FindVMTemplateByTags`, returning a `map[string]int32` (node→vmid) instead of a single node+vmid pair:
  - `localStorage=false` (default): preserves existing behavior — finds exactly one template cluster-wide, respecting TemplateMatchPolicy. If the same tags exist on templates on multiple nodes (e.g. local copies), provisioning fails with "found N VM templates" — use localStorage=true in that scenario.
  - `localStorage=true`: finds one matching template per node in `allowedNodes`. Integrate with the TemplateMatchPolicy feature.
- Scheduler updated to accept the template map and return the selected templateID alongside the target node
- `createVM` reworked: resolves template map first (fail fast), then schedules, then sets `options.Node = options.Target` so the clone runs on the same node that holds the template, avoiding cross-node data transfer

**Testing performed:**

Unit tests:

- All existing unit tests pass (make test)
- New unit tests for FindVMTemplatesByTags covering localStorage=true: all nodes have template, missing template on one node, duplicate templates on a node, empty tag list
- Scheduler and vmservice tests updated for new function signatures

E2E — tested against a 2-node Proxmox VE cluster (pve-1, pve-3) with local-zfs storage, templateSelector with tags `[k8s, v1.33.12, ubuntu-24.04]`  and `localStorage=true`:

- `localStorage=false` + same tags on multiple nodes: provisioning failed with "VM template not found: found 2 VM templates with tags 'k8s;ubuntu-24.04;v1.33.12'". Expected behavior, localStorage=true is required in this setup.
- `localStorage=true` + same tag on multiple nodes: Control plane provisioned on pve-1, worker on pve-3. Each VM cloned from its node-local template with no cross-node data transfer.


api/v1alpha1/proxmoxmachine_types.go
```diff
@@ -202,6 +202,12 @@ type TemplateSource struct {
 	// TemplateSelector defines MatchTags for looking up VM templates.
 	// +optional
 	TemplateSelector *TemplateSelector `json:"templateSelector,omitempty"`
+
+	// LocalStorage indicates that the VM template is stored on local (non-shared) storage.
+	// When true, each node in allowedNodes must have a local copy of the template.
+	// +kubebuilder:default=false
+	// +optional
+	LocalStorage *bool `json:"localStorage,omitempty"`
```
mcbenjemaa: There is no need to add the field in the v1alpha1 version only v1alpha2 applies

chrodriguez: I've kept it because of the @Atoms patched version I used in my lab. I can remove it if you prefer.

mcbenjemaa: New features are only supposed to go to the current api version.

V1alpha1 is only left prior to the  shift towards v1beta1, then it will be deprecated.

chrodriguez: Done, I've just pushed modified code without touching v1alpha1

chrodriguez:
> ## [Quality Gate Failed](https://sonarcloud.io/dashboard?id=ionos-cloud_cluster-api-provider-proxmox&pullRequest=784) **Quality Gate failed**
> Failed conditions  [3 New Critical Issues](https://sonarcloud.io/dashboard?id=ionos-cloud_cluster-api-provider-proxmox&pullRequest=784) (required ≤ 0)
> 
> [See analysis details on SonarQube Cloud](https://sonarcloud.io/dashboard?id=ionos-cloud_cluster-api-provider-proxmox&pullRequest=784)
> 
>  Catch issues before they fail your Quality Gate with our IDE extension  [SonarQube for IDE](https://www.sonarsource.com/products/sonarlint/features/connected-mode/?referrer=pull-request)

I can't see wich rules are failing. How can I fix this 3 critical issues?

chrodriguez:
Hi, I'd like to know how to find the failing SonarCloud checks so I can fix them. Thanks in advance

wikkyk:
@chrodriguez sonar outputs inline annotations: https://github.com/ionos-cloud/cluster-api-provider-proxmox/pull/784/checks?check_run_id=79738687869

The conversion function naming is a known false-positive with sonar and you can ignore it. We tend to just accept error message string duplication (it's a bit silly, deduplicating those makes debugging _harder_ and there's nothing wrong with error messages that happen to be the same right now drifting apart).
The function complexity ones should be taken care of, though.

pkg/proxmox/goproxmox/api_client.go
```diff
@@ -36,6 +35,9 @@ import (
 
 var _ capmox.Client = &APIClient{}
 
+// errFmtCannotGetClusterStatus is the error format used when fetching the Proxmox cluster status fails.
+const errFmtCannotGetClusterStatus = "cannot get cluster status: %w"
```
wikkyk: I specifically asked to keep the repeated error format.

api/v1alpha1/proxmoxmachine_conversion.go
```diff
@@ -168,3 +174,10 @@ func restoreProxmoxMachineSpec(src *ProxmoxMachineSpec, dst *v1alpha2.ProxmoxMac
 		}
 	}
 }
+
+// Convert_v1alpha2_TemplateSource_To_v1alpha1_TemplateSource converts v1alpha2 TemplateSource
+// to v1alpha1. The localStorage field does not exist in v1alpha1 and is intentionally dropped;
+// it is preserved via MarshalData annotation for round-trip fidelity.
+func Convert_v1alpha2_TemplateSource_To_v1alpha1_TemplateSource(in *v1alpha2.TemplateSource, out *TemplateSource, s apiconversion.Scope) error { //NOSONAR name with underscores is required by conversion-gen
```
wikkyk: `//NOSONAR` just suppresses all findings on a line. We prefer to manually review issues.

wikkyk:
Please follow the AI agent attribution rules.

pkg/proxmox/goproxmox/api_client.go
chrodriguez: Reverted

api/v1alpha1/proxmoxmachine_conversion.go
chrodriguez: Reverted

chrodriguez:
> Please follow the AI agent attribution rules.

Thanks for the heads-up. I use an AI assistant for parts of this and should have attributed it. I'll rewrite the commits with a `Co-authored-by:` trailer for the agent and force-push.

chrodriguez:
> Please follow the AI agent attribution rules.

I've just updated the commits with `Co-authored-by:`. Please let me know if it's OK. Thanks!

pkg/proxmox/goproxmox/api_client.go
```diff
@@ -145,8 +144,60 @@ func (c *APIClient) FindVMResource(ctx context.Context, vmID uint64) (*proxmox.C
...
+	}
+	node, vmid, err := c.findSingleVMTemplateByTags(ctx, templateTags, matchPolicy)
+	if err != nil {
+		return nil, err
+	}
+	return map[string]int32{node: vmid}, nil
+}
+
+// normalizeTags lowercases, sorts and deduplicates the given tags in place.
+func normalizeTags(tags []string) []string {
```
wikkyk: This func should also make sure that the tags conform to the permitted character set.

chrodriguez: Fixed in 49ff474 — tags are now validated against the permitted character set (same pattern the CRD enforces) and the lookup returns ErrInvalidTag on violations. I also took the chance to validate matchPolicy and the allowed-nodes input in the local storage path in the same commit.

pkg/proxmox/goproxmox/api_client.go
```diff
@@ -163,13 +168,34 @@ func (c *APIClient) FindVMTemplatesByTags(ctx context.Context, templateTags []st
...
+	case "":
+		return string(infrav1.TemplateMatchPolicyExact), nil
+	default:
+		return "", fmt.Errorf("%w: %q", ErrInvalidMatchPolicy, matchPolicy)
+	}
+}
+
+// permittedTagRegex matches the character set Proxmox allows for tags. It is the
+// lowercase equivalent of the CRD validation pattern on tag fields.
+var permittedTagRegex = regexp.MustCompile(`^[a-z0-9_][a-z0-9_\-+.]*$`)
```
wikkyk: This does not need to be a global.

chrodriguez: Done!

wikkyk:
Just a note to remember that the final merge will need to attribute @Atoms.

chrodriguez:
Is it ok to update branch from main?

theoraim:
Hi ! Do you guys know when this feature will be available ? For my use case, it would be very useful :)

Atoms:
@chrodriguez thanks for taking care of this, great job. 

As i changed company i don't have the need for this and don't have place where to test.

theoraim:
If needed, I can test this PR :)

chrodriguez:
> If needed, I can test this PR :)

Yes, of course. Thanx

api/v1alpha1/proxmoxmachine_conversion.go
```diff
@@ -195,10 +167,17 @@ func restoreProxmoxMachineSpec(src *ProxmoxMachineSpec, dst *v1alpha2.ProxmoxMac
...
+				DefaultIPv6: ptr.To(true),
+				Bridge:      ptr.To("vmbr0"),
 			}},
 		}
 	}
 }
+
+// Convert_v1alpha2_TemplateSource_To_v1alpha1_TemplateSource converts v1alpha2 TemplateSource
+// to v1alpha1. The localStorage field does not exist in v1alpha1 and is intentionally dropped;
+// it is preserved via MarshalData annotation for round-trip fidelity.
```
wikkyk, line 180: What's up with this comment?

wikkyk:
This PR will conflict with #720 because they both introduce a concept of 'effective allowedNodes', just incompatibly.

api/v1alpha1/proxmoxmachine_conversion.go
chrodriguez: This wrapper is required by conversion-gen: since `localStorage` has no v1alpha1 counterpart, the generator only emits the private `autoConvert_...` and expects the project to provide the public `Convert_...` function manually — the generated code in `zz_generated.conversion.go` registers and calls it by that exact name, so it can't be removed or renamed.

The comment is there to document that dropping `localStorage` on down-conversion is intentional and not lossy: `ConvertFrom` stores the full hub object in the conversion annotation via `MarshalData`, and `ConvertTo` restores the field from it in `restoreProxmoxMachineSpec`. Happy to reword it if it reads confusing — e.g. leading with the conversion-gen requirement instead of the field semantics.

For reference, these are the generated call sites that depend on it:

- [`api/v1alpha1/zz_generated.conversion.go:1661`](https://github.com/Mikroways/cluster-api-provider-proxmox/blob/ea540281968e5d00124f1bf7c47fd94e99f505e0/api/v1alpha1/zz_generated.conversion.go#L1661) — the generated `autoConvert` carries the marker conversion-gen leaves when it refuses to emit the public function:

  ```go
  // WARNING: in.LocalStorage requires manual conversion: does not exist in peer-type
  ```

- [`api/v1alpha1/zz_generated.conversion.go:428`](https://github.com/Mikroways/cluster-api-provider-proxmox/blob/ea540281968e5d00124f1bf7c47fd94e99f505e0/api/v1alpha1/zz_generated.conversion.go#L428) — `RegisterConversions` registers the function with the scheme:

  ```go
  return Convert_v1alpha2_TemplateSource_To_v1alpha1_TemplateSource(a.(*v1alpha2.TemplateSource), b.(*TemplateSource), scope)
  ```

- [`api/v1alpha1/zz_generated.conversion.go:1784`](https://github.com/Mikroways/cluster-api-provider-proxmox/blob/ea540281968e5d00124f1bf7c47fd94e99f505e0/api/v1alpha1/zz_generated.conversion.go#L1784) — the `VirtualMachineCloneSpec` conversion calls it directly:

  ```go
  if err := Convert_v1alpha2_TemplateSource_To_v1alpha1_TemplateSource(&in.TemplateSource, &out.TemplateSource, s); err != nil {
  ```

Removing the function fails the build with `undefined: Convert_v1alpha2_TemplateSource_To_v1alpha1_TemplateSource` at the last two sites, and regenerating doesn't change that: as long as the field exists only in v1alpha2, conversion-gen will always require this manual function. The round-trip preservation lives in [`proxmoxmachine_conversion.go`](https://github.com/Mikroways/cluster-api-provider-proxmox/blob/ea540281968e5d00124f1bf7c47fd94e99f505e0/api/v1alpha1/proxmoxmachine_conversion.go#L113) (`MarshalData` in `ConvertFrom`, restore at lines [113-116](https://github.com/Mikroways/cluster-api-provider-proxmox/blob/ea540281968e5d00124f1bf7c47fd94e99f505e0/api/v1alpha1/proxmoxmachine_conversion.go#L113)), same pattern as the existing manual `Convert_*` functions in `api/v1alpha1/conversion.go`.

chrodriguez:
> This PR will conflict with #720 because they both introduce a concept of 'effective allowedNodes', just incompatibly.

Good point. Looking at #720, the overlap is the resolution of "which allowedNodes apply to this machine": #720 does it via `MachineScope.GetEffectiveAllowedNodes()` (failure domain → machine spec, with the cluster fallback at each call site), while this PR extracted a `resolveAllowedNodes` helper in vmservice (machine spec → cluster) that feeds both the template lookup and the scheduler.

To make the two converge, I can move my helper onto `MachineScope` as `GetEffectiveAllowedNodes()` with the cluster fallback folded in, matching #720's shape. Then whichever PR lands second only needs to add its layer inside that single method — and the local storage template lookup will automatically honor failure domains, since it already receives the resolved node list. Happy to do that here if it helps, or to rebase on top of #720 if it merges first.

chrodriguez:
> > This PR will conflict with #720 because they both introduce a concept of 'effective allowedNodes', just incompatibly.
> 
> Good point. Looking at #720, the overlap is the resolution of "which allowedNodes apply to this machine": #720 does it via `MachineScope.GetEffectiveAllowedNodes()` (failure domain → machine spec, with the cluster fallback at each call site), while this PR extracted a `resolveAllowedNodes` helper in vmservice (machine spec → cluster) that feeds both the template lookup and the scheduler.
> 
> To make the two converge, I can move my helper onto `MachineScope` as `GetEffectiveAllowedNodes()` with the cluster fallback folded in, matching #720's shape. Then whichever PR lands second only needs to add its layer inside that single method — and the local storage template lookup will automatically honor failure domains, since it already receives the resolved node list. Happy to do that here if it helps, or to rebase on top of #720 if it merges first.

Is it OK to proceed as mentioned above, @wikkyk ?

emanuelebosetti:
Author of #720 here. After the last review round it resolves placement once in the MachineScope and exposes it as `machineScope.AllowedNodes()`, which is pretty much the single method @chrodriguez suggested above.

I tried merging this branch on top of #720 locally: two small conflicts, both resolved by passing `machineScope.AllowedNodes()` to the new `ScheduleVM` signature, and the local-storage template lookup picks up failure domain support for free. Tests green. Details in #720.

chrodriguez:
Thanks for aligning `AllowedNodes()` with the single-method approach — exactly what I had in mind.

Quick check: would basing #784 on top of #720 fully resolve the overlap? From your description, once `machineScope.AllowedNodes()` exists, my `resolveAllowedNodes` helper goes away, the local-storage lookup picks up failure-domain support for free, and the only change is passing `AllowedNodes()` into the new `ScheduleVM` signature.

If so, I'd rather not merge #720 into this PR (it pulls in the whole branch as a dependency). Would it work to keep #784 self-contained and do the trivial swap once #720 lands?

PR #791, opened 2026-06-14
Proxmox HA support for ProxmoxMachines: register on create, delete with purge (fixes #216)

### LINKED ISSUE #216: The ability for the CAPMOX provider to work with HA
We have discovered that the CAPMOX provider cannot delete machines which are part of high availability.
Steps to reproduce:

Step 1. VM128 which is a worker node for a cluster is added into High Availability in the proxmox GUI (Datacenter > High Availability)

Step 2. Find the machine resource in the management cluster and delete it

kubectl get machines|grep mk1-busi-cl-worker-wrn46
kubectl delete machine mk1-busi-cl-workers-lnlc8-q98l6 

Step 3 - Check the CAPMOX provider logs
E0515 09:29:06.758515       1 controller.go:329] "Reconciler error" err="cannot delete vm with id 128: 500 unable to remove VM 128 - used in HA resources and purge parameter not set." controller="proxmoxmachine" controllerGroup="infrastructure.cluster.x-k8s.io" controllerKind="ProxmoxMachine" ProxmoxMachine="default/mk1-busi-cl-worker-wrn46" namespace="default" name="mk1-busi-cl-worker-wrn46" reconcileID="6631f758-0c64-4776-9f4a-aadf0435510f"

I would like to request that HA can be added to the proxmox machine template and that if HA is added the machine is added to HA.

When deleting a machine with HA the purge parameter can be added to delete machines in HA.

There is a purge parameter for the DELETE API request.
[https://pve.proxmox.com/pve-docs/api-viewer/index.html#/nodes/{node}/qemu/{vmid}](https://pve.proxmox.com/pve-docs/api-viewer/index.html#/nodes/%7Bnode%7D/qemu/%7Bvmid%7D)

Environment:

Cluster-api-provider-proxmox version: v1.6.3/v0.40
Kubernetes version: (use kubectl version): 1.28.8
OS (e.g. from /etc/os-release): ubuntu 2204

ibra-coul:
## What

Implements both halves of #216: opt-in Proxmox High Availability (HA) for the VM backing a `ProxmoxMachine`.

### 1. Delete HA-managed VMs (the reported bug)

A `ProxmoxMachine` whose VM had been added to a Proxmox HA group could not be deleted — PVE rejects the plain `DELETE /nodes/{node}/qemu/{vmid}` with:

```
cannot delete vm with id 128: 500 unable to remove VM 128 - used in HA resources and purge parameter not set.
```

…leaving the `Machine` stuck in `Deleting` with the finalizer never removed.

`DeleteVM` now takes an explicit `purge bool`. The vmservice caller passes `purge=true` when the machine opted into HA (`Spec.HighAvailability.Enabled`); on a delete rejection the VM is then deleted again with `purge=1`, which makes PVE drop the HA reference (plus any replication / backup-job references). The purge delete is issued through the version-stable `Client.Delete` (`purge=1` in the query string), and the cloud-init ISO cleanup done on the first attempt is preserved. Unrelated delete failures are surfaced unchanged.

### 2. Register VMs in HA on create (the feature request)

New `ProxmoxMachineSpec.highAvailability { enabled, state }`, inherited by `ProxmoxMachineTemplate` so an entire MachineDeployment / control plane can opt in. When `enabled`, the VM is registered as a Proxmox HA resource once it is `Ready` (from `vmservice.ReconcileVM`, alongside the rest of the VM lifecycle), via the new idempotent `Client.EnsureHAResource` (create-or-update of `/cluster/ha/resources`). It is re-asserted on every reconcile, so it self-heals if the resource is removed out-of-band. `state` is one of `started|stopped|disabled|ignored` (default `started`).

## Notes

- HA is managed through `/cluster/ha/resources` on every supported PVE version; the PVE 9 migration from HA *groups* to HA *rules* did not change this endpoint, so no version-specific handling is needed.
- `HighAvailability` has no v1alpha1 equivalent: a manual `Convert_v1alpha2_ProxmoxMachineSpec_To_v1alpha1_ProxmoxMachineSpec` drops it on down-conversion and it is restored from the Hub-data annotation on up-conversion, so v1alpha1 round-trips stay lossless (covered by the conversion fuzz tests).
- Tested end-to-end on **Proxmox VE 9.2.2** (go-proxmox v0.6.0): delete of an HA-managed VM, registration on create via the spec field, and rollout via `ProxmoxMachineTemplate`.
- Unit tests added for `DeleteVM` (purge paths), `EnsureHAResource` (create / update / no-op), HA registration in `ReconcileVM`, and purge-on-HA delete in vmservice.
- Generated files (CRDs, deepcopy, conversion, mocks) regenerated with controller-gen + conversion-gen + mockery.

Fixes #216


pkg/proxmox/goproxmox/api_client.go
```diff
@@ -249,12 +249,94 @@ func (c *APIClient) DeleteVM(ctx context.Context, nodeName string, vmID int64) (
...
 	return task, nil
 }
 
+// deleteVMWithPurge deletes a VM passing purge=1, which instructs Proxmox to
+// also remove the VM from any HA, replication and backup-job configuration it
+// is referenced by. go-proxmox's VirtualMachine.Delete does not expose the
+// purge parameter, so the request is issued directly against the API.
+func (c *APIClient) deleteVMWithPurge(ctx context.Context, vm *proxmox.VirtualMachine) (*proxmox.Task, error) {
+	var upid proxmox.UPID
+	if err := c.DeleteWithParams(ctx,
```
timtasse: the func is called only Delete, it was a api breaking change
and you need to update the dependency for this library to 0.8.0 to use it.

pkg/proxmox/goproxmox/api_client.go
```diff
@@ -249,12 +249,94 @@ func (c *APIClient) DeleteVM(ctx context.Context, nodeName string, vmID int64) (
...
+// this endpoint), so no version-specific handling is required here.
+func (c *APIClient) EnsureHAResource(ctx context.Context, vmID int64, state string) error {
+	if state == "" {
+		state = "started"
+	}
+	sid := fmt.Sprintf("vm:%d", vmID)
+
+	var resources []struct {
+		SID   string `json:"sid"`
+		State string `json:"state"`
```
timtasse, line 278: in PVE 9.2 there is a new optional feature "auto-rebalance" and it is enabled by default.
i think it could be smart to implement it with false, so every VM is located where it was created and only in case of HA occurred, it runs on another host.
i don't know if you can set this parameter on an older PVE version or if it creates an error.

internal/controller/proxmoxmachine_controller.go
```diff
@@ -235,6 +235,18 @@ func (r *ProxmoxMachineReconciler) reconcileNormal(ctx context.Context, machineS
 		return reconcile.Result{RequeueAfter: infrav1.DefaultReconcilerRequeue}, nil
 	}
 
+	// If the machine opted into Proxmox HA, register the VM as an HA resource
+	// now that it is provisioned. This is idempotent and re-asserted on every
+	// reconcile, so it also recovers HA membership if it is removed out-of-band.
+	if ha := machineScope.ProxmoxMachine.Spec.HighAvailability; ha != nil && ha.Enabled {
```
mcbenjemaa: Since this is affecting the VM in Proxmox, it's better to place this block in the `ReconcileVM()` function.

pkg/proxmox/goproxmox/api_client.go
```diff
@@ -249,12 +249,94 @@ func (c *APIClient) DeleteVM(ctx context.Context, nodeName string, vmID int64) (
...
+		// The cloud-init ISO cleanup that vm.Delete performs already ran during
+		// the attempt above (it happens before the failing DELETE call), so the
+		// purge retry only needs to issue the delete itself.
+		//
+		// See https://github.com/ionos-cloud/cluster-api-provider-proxmox/issues/216
+		if !isUsedInHAError(err) && vm.HA.Managed != 1 {
+			return nil, fmt.Errorf("cannot delete vm with id %d: %w", vmID, err)
+		}
+
+		task, err = c.deleteVMWithPurge(ctx, vm)
```
mcbenjemaa: There is no need to place this Call here,
Make the `DeleteVM()` function accept a bool argument `purge`.

ibra-coul:
Okay, noted. I'll fix it today.

pkg/proxmox/goproxmox/api_client.go
ibra-coul: Done in f1f0851. `DeleteVM` now takes an explicit `purge bool`; the HA-detection heuristics (the `ha.managed` flag and the `used in HA resources` error-string match) are gone. The vmservice caller passes `purge=true` when the machine opted into HA (`Spec.HighAvailability.Enabled`).

pkg/proxmox/goproxmox/api_client.go
ibra-coul: Fixed in f1f0851. The purge delete now goes through the version-stable `Client.Delete` with `purge=1` in the query string instead of `DeleteWithParams`, so no go-proxmox bump is needed (this PR is based on the v0.6.0 bump already on main, which has `Delete`).

internal/controller/proxmoxmachine_controller.go
ibra-coul: Done in f1f0851. The HA registration block moved from the controller into `vmservice.ReconcileVM`, next to the rest of the VM lifecycle; it runs once the VM is provisioned and is still re-asserted on every reconcile.

pkg/proxmox/goproxmox/api_client.go
ibra-coul: Deferred for now (not in f1f0851). `auto-rebalance` is a PVE 9 datacenter HA setting and, as you note, it is unclear whether passing it on `/cluster/ha/resources` errors on older PVE. Since #216 is only about registering on create / purging on delete, I would rather keep the default behaviour here and add `auto-rebalance=false` as a follow-up once it can be tested against both PVE 9.2 and an older release. Happy to fold it in if you prefer it in this PR.

pkg/proxmox/goproxmox/api_client.go
timtasse: why let you produce the error in the first place and then in a second run set the purge=1 flag, this would spam the log of the proxmox server. at this moment you are sure the VM should be deleted with purge, so directly use it with purge, would be a better approach in my opinion.

pkg/proxmox/goproxmox/api_client.go
timtasse: atm i have no proxmox version prior 9.2, so i can actually not test this, but yes we can do this in a separate pull request.

pkg/proxmox/goproxmox/api_client.go
timtasse: i see, you use the underlying function and not the changed one.
i don't know if this is the best option, but it works for sure.
the changed func uses a new struct and so supports all parameters in a nicer and typo save way.

pkg/proxmox/goproxmox/api_client.go
ibra-coul: Good point — done in 07ff856. When purge is requested the VM is known to be HA-managed, so `DeleteVM` now issues the purge=1 delete directly; no plain delete is attempted first, so the PVE task log is no longer spammed with the "used in HA resources" rejection.

pkg/proxmox/goproxmox/api_client.go
wikkyk: > i don't know if this is the best option, but it works for sure.

It's not.

api/v1alpha1/proxmoxmachine_conversion.go
```diff
@@ -109,6 +109,12 @@ func restoreProxmoxMachineSpec(src *ProxmoxMachineSpec, dst *v1alpha2.ProxmoxMac
 
 	Convert_string_To_Pointer_string(src.TemplateSource.SourceNode, ok, restored.TemplateSource.SourceNode, &dst.TemplateSource.SourceNode)
 
+	// HighAvailability does not exist in v1alpha1; restore it from the Hub
+	// annotation so a v1alpha1 round-trip does not drop HA configuration.
```
wikkyk: This comment is unnecessary.

api/v1alpha2/proxmoxmachine_types.go
```diff
@@ -139,6 +139,31 @@ type ProxmoxMachineSpec struct {
...
+
+// HighAvailabilitySpec configures Proxmox High Availability for a VM.
+type HighAvailabilitySpec struct {
+	// enabled registers the VM as a Proxmox HA resource once it is provisioned.
+	// +kubebuilder:default=false
+	Enabled bool `json:"enabled"`
+
+	// state is the requested HA state for the resource, mapped to the Proxmox
+	// HA resource "state" (request state). Defaults to "started".
+	// +kubebuilder:validation:Enum=started;stopped;disabled;ignored
```
wikkyk: I would like to see consts for these. Ideally, this would be a real enum type.

pkg/proxmox/goproxmox/api_client.go
```diff
@@ -255,6 +274,63 @@ func (c *APIClient) DeleteVM(ctx context.Context, nodeName string, vmID int64) (
...
 
+// deleteVMWithPurge deletes a VM passing purge=1, which instructs Proxmox to
+// also remove the VM from any HA, replication and backup-job configuration it
+// is referenced by. go-proxmox's VirtualMachine.Delete does not expose the
+// purge parameter, so the request is issued directly against the API. Proxmox
+// DELETE endpoints take options as query parameters, so purge=1 is appended to
+// the path and the request goes through the version-stable Client.Delete.
+func (c *APIClient) deleteVMWithPurge(ctx context.Context, vm *proxmox.VirtualMachine) (*proxmox.Task, error) {
+	var upid proxmox.UPID
+	path := fmt.Sprintf("/nodes/%s/qemu/%d?purge=1", vm.Node, vm.VMID)
```
wikkyk: Do this with `VirtualMachineDeleteOptions` as per https://github.com/luthermonson/go-proxmox/blob/v0.8.0/migration/v0.8.0.md

pkg/proxmox/goproxmox/api_client.go
```diff
@@ -212,8 +212,13 @@ NEXT_VM:
 	return vmTemplate.Node, int32(vmTemplate.VMID), nil
 }
 
-// DeleteVM deletes a VM based on the nodeName and vmID.
-func (c *APIClient) DeleteVM(ctx context.Context, nodeName string, vmID int64) (*proxmox.Task, error) {
+// DeleteVM deletes a VM based on the nodeName and vmID. When purge is true the
+// VM is deleted with purge=1, which instructs Proxmox to drop the VM from any
+// HA, replication and backup-job configuration it is referenced by. Callers that
+// register VMs as Proxmox HA resources must pass purge=true, since PVE otherwise
```
wikkyk: Since we know that a VM is a HA resource, we don't need the purge param. `DeleteVM` can figure out how to delete the VM properly by itself.

pkg/proxmox/goproxmox/api_client.go
ibra-coul: Done in 208640a. Bumped go-proxmox to v0.8.0 and switched to `vm.Delete(ctx, &proxmox.VirtualMachineDeleteOptions{Purge: true})`.

pkg/proxmox/goproxmox/api_client.go
ibra-coul: You were right. The hand-rolled request also skipped go-proxmox's `deleteCloudInitISO`, which `vm.Delete` does for you, so HA VMs were leaving their cloud-init ISO behind on the storage. Gone now that it goes through `vm.Delete` with the options struct.

pkg/proxmox/goproxmox/api_client.go
ibra-coul: Done. `DeleteVM(ctx, node, vmID)` now always purges. Purge only drops the VM from HA / replication / backup-job config, which is what we want whenever we destroy a VM we own, and it also works if the VM was put into HA outside of the spec.

api/v1alpha2/proxmoxmachine_types.go
ibra-coul: Done. Added a `HighAvailabilityState` type with consts for the four states; it is used from the CRD down to the Proxmox client.

api/v1alpha1/proxmoxmachine_conversion.go
ibra-coul: Removed.

api/v1alpha2/proxmoxmachine_types.go
```diff
@@ -139,6 +139,51 @@ type ProxmoxMachineSpec struct {
...
+type HighAvailabilitySpec struct {
+	// enabled registers the VM as a Proxmox HA resource once it is provisioned.
+	// Defaults to false when not specified.
+	// +kubebuilder:default=false
+	// +optional
+	Enabled *bool `json:"enabled,omitempty"`
+
+	// state is the requested HA state for the resource, mapped to the Proxmox
+	// HA resource "state" (request state). Defaults to "started".
+	// +kubebuilder:default=started
```
wikkyk, line 184: I am not certain if `started` is the correct default.

That being said, perhaps it is the correct default but it's the name of the field that's wrong. It's not a _state_, it's a _request_.

wikkyk (commented):
The mechanism is reasonable but I think the word 'state' isn't a good fit for the field, type, and consts. 'State' suggests this is the actual running state, when in reality this is a _request_ for Proxmox.

api/v1alpha2/proxmoxmachine_types.go
ibra-coul: > I am not certain if `started` is the correct default.

Checked the power paths: `reconcilePowerState` only ever powers **on**, and the only `.Stop(` in the non-test tree is in `DeleteVM`, right before destroying the VM. HA is registered last in `ReconcileVM`, on an already-running VM. Nothing in the provider wants a VM to stay off, so `started` can't fight the HA manager, and it matches PVE's own default.

> 'State' suggests this is the actual running state

Agreed. Proposal: `requestedState` / `HighAvailabilityRequestedState`. Plain `request` works for me too, your call.

ibra-coul:
@wikkyk gentle ping on the last open thread: the `state` vs `request` naming is the only thing left. Everything else is resolved and CI is green.

To keep this moving: unless you'd rather have something else, I'll go with `requestedState` (field `highAvailability.requestedState`, type `HighAvailabilityRequestedState`) and push the rename in the next couple of days. If plain `request` reads better to you, just say so and I'll use that instead.

This has been open since June and the change is small and self-contained.

PR #792, opened 2026-06-14
Refactor routingdata safety

65278:
*Issue #, if available:*

*Description of changes:*
This is a large refactor that spun from the uniq gateway hotfix. I realised that this introduced a couple of subtle bugs, which lead to refactors. Refactors lead to more refactors, and I stopped at this point. I made two subsequent issues, and am considering a third one about moving metadata rendering out of cloudinit.

Changes in order:
 - type safe RoutingData and FIBRuleData. Validation of Prefixes/Addresses happens in bootstrap, not in render.
 - Gateways are now RoutingData and on longer special cases.
 - Extended route collision detection to all routes, not just default gateways.
 - Is6 now differentiates Routes/Rules which only consist of placeholders: (`default`|`all`)
 - `pkg/types` is not `pkg/network` because Network is now an interface.
 - This allows cloudinit and ignition to share one validator.
 - This also allows implementation of new renderers which implement a subset of cloudinit/ignition (netplan v1 was requested)

*Testing performed:*
v1alpha2 clusters rolled out with loadbalancers for ignition and cloudinit.


65278:
I'm not really willed to fix kolmogorov complexity here. At least for Validate(), this really can't be helped. The other branch I might refactor, but since every `if foo, err := bar(); {}`  counts, I really think it's a useless metric for golang as implemented.

wikkyk:
My only real criticism is that I would've made the placeholders explicitly state IP version instead of adding the Is6 field. However, that could be considered an API change so we could do that in v1alpha3. The conversion should be straightforward.

pkg/ignition/network.go
```diff
@@ -179,19 +231,25 @@ func render(name string, tpl string, data types.NetworkConfigData) ([]byte, erro
...
+			for j := range data {
+				if data[j].Name == child {
+					data[j].VRF = data[i].Name
+					break
 				}
 			}
 		}
+		// adjust VRF routes by adding the routing table to each member route.
+		// This is to keep approximate expected behaviour with netplan v2, but
+		// could use its own validation pass in the future.
```
wikkyk, line 250: I think this comment's tone isn't serious enough. We validate data (spec tables) that is then potentially discarded and replaced with other data (vrf tables). So what we render isn't necessarily what we've validated. This only applies when you have both an explicit route.Table and a vrf table that are divergent, which is arguably a config error but we don't catch that.

wikkyk: Also, 

>adjust VRF routes by adding the routing table to each member route.

is somewhat ambiguous as to which routing table is added and to what.

wikkyk (approved):

Issue #758, opened 2026-05-18
ProxmoxMachine does not recover IPAM/address state after Velero restore; allocated IPAddress exists but VM remains WaitingForStaticIPAllocation


Leonorus:
**What steps did you take and what happened:**

We restored a CAPI management cluster from Velero backup while the workload cluster VMs/nodes still existed in Proxmox and the workload Kubernetes cluster itself was mostly healthy.

After the restore, capmox/CAPI/IPAM status was not fully reconstructed from the already-existing workload state and restored IPAM objects. We then hit two related failure modes:

1. Existing restored Machines lost address/status information even though the workload Nodes were Ready and correlated correctly.
2. A later replacement ProxmoxMachine got stuck in `WaitingForStaticIPAllocation` even though the matching `IPAddressClaim` and `IPAddress` already existed and the claim had `status.addressRef` pointing to the allocated address.

This does not look like a simple case of IPAM failing to allocate an address. In the severe rollout case, IPAM had allocated the address and the Kubernetes IPAM objects existed. The problem is that capmox did not reliably consume/recover that existing IPAM state after restore/reconcile, so it did not populate ProxmoxMachine IP status, did not tag the Proxmox VM with the allocated IP, and did not start the VM.

### 1. Original post-Velero-restore symptom: workload was healthy, management/status plane was degraded

After Velero restore, the workload cluster was often healthy from the workload side:

```text
workload Kubernetes /readyz: ok
workload Nodes: Ready
Talos etcd: healthy on the current control-plane members
workload Node providerIDs: matched CAPI Machines
workload Node InternalIPs: present
```

But the restored management/status plane had missing capmox/CAPI address state:

```text
Machine.status.addresses: empty
ProxmoxMachine.status.addresses: empty
ProxmoxMachine.status.ipAddresses: empty
```

We also observed IPAM restore drift:

```text
many restored IPAddressClaim.metadata.ownerReferences: []
some old IPAM objects remained for Machines that no longer existed
some cases had an IPAddress object while the corresponding IPAddressClaim was missing
```

This caused TalosControlPlane/CABPT to report failures even when the actual workload cluster was healthy:

```text
ControlPlaneComponentsHealthy=False/ControlPlaneComponentsInspectionFailed
machine "...": no addresses were found for node "..."

EtcdClusterHealthyCondition=False/EtcdClusterUnhealthy
error checking etcd health on machine "...": no addresses were found for node "..."

[...]

PR #799, opened 2026-06-18
Issue 758 status reconstruction

Leonorus:
**Issue  #758

**Description of changes:**

After a Velero restore of a CAPMOX management cluster, already-running workload VMs can lose their address state: `ProxmoxMachine.status.ipAddresses` and `status.addresses` come back empty even though the VM, Node, and the `IPAddressClaim`/`IPAddress` objects are all healthy. This leaves CAPI `Machine.status.addresses` empty and breaks downstream controllers — e.g. TalosControlPlane/CABPT fail with `no addresses were found for node`.

The root cause is that `ReconcileVM`'s reason-gated state machine never republishes status for these machines. For an already-running VM, `ensureVirtualMachine` leaves the `VirtualMachineProvisioned` reason at `Cloning`, and `reconcileVirtualMachineConfig` early-returns because the VM is running, so the reason never advances to `WaitingForStaticIPAllocation`. As a result `reconcileIPAddresses` and `reconcileMachineAddresses` are never reached and status stays empty.

This PR adds `reconcileAddressRecovery`, a read-only, idempotent step run right after `ensureVirtualMachine`. When the VM is running and `status.ipAddresses` is empty, it reconstructs status purely from the existing IPAM objects (reusing the claim-rooted `ResolveIPAddressClaim` from the first PR) and republishes `status.ipAddresses` and `status.addresses`. It:

- never creates `IPAddressClaim`/`IPAddress` objects, mutates the Proxmox VM, or advances the provisioning condition;
- republishes only when **every** expected claim resolves cleanly (`ClaimResolved`) — orphaned or conflicting IPAM objects are deliberately left to the regular provisioning path / follow-up PRs;
- is a no-op during normal provisioning (a freshly cloned VM is not running at the IP stage, and once status is populated it never fires again).

Supporting refactors (behaviour unchanged for the provisioning path): the shared device/pool walk is extracted into `collectDeviceAddresses(ctx, scope, addresses, resolve)` with a pluggable resolver (`handleIPAddresses` for provisioning, `resolveExistingIPAddress` read-only for recovery), and the status-write loop is extracted into `writeIPAddressStatus`, reused by both paths.

**Testing performed:**

- `go test ./internal/service/vmservice/...` → 169 pass, including 6 new `reconcileAddressRecovery` tests (recovers status from existing claims; no-op when VM not running / VM absent / status already populated / `Spec.Network` nil; skips without writing status or creating objects when a claim is missing) and all pre-existing provisioning tests (regression guard for the refactor).
- `go test ./pkg/kubernetes/ipam/...` → 24 pass.
- `go vet` and `gofmt` clean on the changed files.
- `make lint` clean for the changed files.
- `make verify` clean on the committed tree — no generated-file drift, no API changes.

### DIFF
--- a/internal/service/vmservice/recovery_test.go
+++ b/internal/service/vmservice/recovery_test.go
@@ -1,0 +1,176 @@
+/*
+Copyright 2023-2026 IONOS Cloud.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package vmservice
+
+import (
+	"context"
+	"reflect"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+	corev1 "k8s.io/api/core/v1"
+	ipamicv1 "sigs.k8s.io/cluster-api-ipam-provider-in-cluster/api/v1alpha2"
+	clusterv1 "sigs.k8s.io/cluster-api/api/core/v1beta2"
+	"sigs.k8s.io/cluster-api/util/conditions"
+
+	infrav1 "github.com/ionos-cloud/cluster-api-provider-proxmox/api/v1alpha2"
+)
+
+func TestReconcileAddressRecovery_RecoversStatus(t *testing.T) {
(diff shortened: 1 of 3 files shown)


wikkyk:
@Leonorus can you rebase this on the go-proxmox 0.8 update?

Leonorus:
Done. But i can't check sonarqube report

wikkyk:
@Leonorus https://github.com/ionos-cloud/cluster-api-provider-proxmox/pull/799/changes#diff-182abf631a02626930f90fea1c6e5b1c70f2beeb192a45973a5e2be881cf242fR294

The complexity of `collectDeviceAddresses` is too high but the complexity is pre-existing. I wouldn't worry about it. It's just that you touched it, so sonar is complaining about it again :-)

internal/service/vmservice/recovery_test.go
```diff
@@ -0,0 +1,176 @@
+/*
+Copyright 2023-2026 IONOS Cloud.
```
wikkyk:
```suggestion
Copyright 2026 IONOS Cloud.
```

new file :)

Issue #607, opened 2026-02-02
Controlplane nodes InternalIP the kube-vip IP


mcbenjemaa:
**What steps did you take and what happened:**
[A clear and concise description of what the bug is.]

Controlplane nodes are showing the InternalIP as the primary IP for the kube-vip.

**What did you expect to happen:**

Controlplane must select its primary IP instead.

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

**Environment:**

- Proxmox version:  v8.x
- Cluster-api-provider-proxmox version: v0.7
- Kubernetes version: (use `kubectl version`):
- OS (e.g. from `/etc/os-release`):

Issue #790, opened 2026-06-14
Render ignition ini files for systemd-networkd with ini file renderer


65278:
**Describe the solution you'd like**
pkg/ignition/network.go renders systemd-networkd units via text/template (networkConfigTplNetworkd) with hand-managed whitespace trim markers. 
These are a pain to manage and extend.
Proposal: build units from structured data  and serialize, using github.com/coreos/go-systemd/unit (which is already a transitive dependency):
Here's a possible data structure for ignition:
```
  opts := []*unit.UnitOption{
      {Section: "Match",   Name: "MACAddress", Value: mac},
      {Section: "Address", Name: "Address",    Value: a1},
      {Section: "Address", Name: "Address",    Value: a2}, // repeated section: fine
  } 
  b, _ := io.ReadAll(unit.Serialize(opts))
```
We can't really use a generic ini library because systemd has decided to neither implement ini nor toml, instead it has chosen to implement repeat sections, which are undefined in the ini spec.

**Anything else you would like to add:**
Independent of similar work in netplan/cloudinit YAML / #780 and can thus land separately. Therefore I've split the issue.

**Environment:**
Not applicable.

Issue #780, opened 2026-06-02
Replace YAML rendering in netplan with actual yaml.


65278:
**Describe the solution you'd like**
We use go text/template to render yaml for cloudinit/netplan and then validate it through yaml parsing.
This is nonsensical, we can make yaml directly and then marshal it. This approach is cleaner and more flexible.

**Anything else you would like to add:**
Depending on how deep we go, ignition can either live off of the yaml or get its own structured ini creation.

**Environment:**
Does not apply.

PR #825, opened 2026-07-01
feat(cloudinit): inject allocated IPs into NoCloud meta-data

vidarno:
Relates to #607. Control-plane nodes end up advertising the kube-vip VIP as
their `InternalIP` because there is no declarative way to point kubelet's
`--node-ip` at the node's *own* allocated address. Today operators work around
this with `preKubeadmCommands` that scrape `ip addr` output — brittle, and
especially painful on IPv6-primary dual-stack clusters.

*Description of changes:*

Expose the IPv4/IPv6 addresses CAPMOX allocates via IPAM in the cloud-init
NoCloud `meta-data` file, so bootstrap consumers can reference them through the
cloud-init datasource, e.g. `{{ ds.meta_data.ipv6 }}`.

Keys emitted (only when an address is allocated):
- `ipv4`, `ipv4_prefix`, `ipv4_gateway` and the `ipv6` equivalents — the
  cluster default-gateway address (the IP on the device flagged default for
  that family, gateway taken from the device's default route).
- `ipv4_<device>`, `ipv6_<device>` and their `_prefix`/`_gateway` variants —
  one set per network device with an allocated IP, suffixed by the Proxmox
  device name (`net0`, `net1`, …), mirroring the existing `ip_net0_<addr>` tag
  convention.

Single-stack clusters emit no empty `ipv6*` keys; NICs with no IPs (and VRFs)
are skipped. Both the cloud-config and Ignition bootstrap paths share the
metadata renderer, so the keys are available regardless of bootstrap format.

No API changes — injection is unconditional, mirroring the precedent set by
#360 (kubernetes-version injection). The renderer now takes a `MetadataInput`
struct instead of a growing positional argument list (addressing the review
feedback on #360).

*Testing performed:*

- `make lint test verify` — green (unit + envtest).
- New unit tests for `PopulateNetworkAddresses` / metadata rendering, plus
  envtest coverage of the cloud-config and Ignition bootstrap paths asserting
  the injected keys.
- Manually verified on a live dual-stack (IPv4 + IPv6) cluster on CAPMOX
  v0.9.0. On a control-plane node, `/run/cloud-init/instance-data.json`
  `ds.meta_data` contained the expected keys (addresses redacted):

      ipv4: 192.0.2.40            ipv6: 2001:db8::29
      ipv4_prefix: 25             ipv6_prefix: 64
      ipv4_gateway: 192.0.2.1     ipv6_gateway: 2001:db8::1
      ipv4_net0: …                ipv6_net0: …   (+ per-device _prefix/_gateway)

---
This PR was prepared with AI assistance (Claude Code); see the `Co-Authored-By`
trailer on the commit. I have reviewed and am responsible for the change.


vidarno:
The failing `test-go`, `test-sh`, and `sonar` checks aren't code failures — all
three bail at the `actions/checkout` step with:

> Refusing to check out fork pull request code from a 'pull_request_target'
> workflow.

These jobs run under `pull_request_target` (they need repo secrets), so
checkout won't pull fork code into that trusted context until a maintainer
approves the workflow run for this PR. The tests never execute, which is why
the follow-on steps then report `No files were found` for `cover.out` /
`sonarqube.xml`. This matches the note in CONTRIBUTING.md about external PRs
needing approval to run actions.

The safe-context checks that did run all pass (`lint`, `lint-api`,
`build image`, `codespell`, `yamllint`, `actionlint`, `scan`, `pre-screen`).

Locally, `make lint test verify` is green (unit + envtest, incl. the new
cases), and I've verified the feature end-to-end on a live dual-stack cluster.

Could a maintainer approve the workflow run (and set any `e2e/*` labels as
appropriate)? Happy to rebase or adjust if anything's needed.

wikkyk:
This isn't a bad idea at all. However, I would like to have #780 #790 tackled first instead of piling even more stuff onto the unmaintainable templating machinery.

vidarno:
Understandable, I'll start work on #780

Issue #796, opened 2026-06-17
bug: TypedLocalObjectReference used as map key compares pointer identity not string value


wikkyk:
Found by copilot and described by claude:

## Summary

`handleDevices` in `internal/service/vmservice/ip.go` uses `corev1.TypedLocalObjectReference` as a map key in `poolMap` and `defaultPoolMap`. Because `TypedLocalObjectReference.APIGroup` is a `*string`, Go's map key equality compares pointer identity rather than the pointed-to string value.

When pool refs are deserialized from YAML (the normal runtime path), each occurrence allocates a fresh `*string` for `APIGroup`. Two structurally identical pool refs will therefore hash to different keys, causing the address accumulation logic to create separate map entries instead of appending to the same one. This silently defeats the multi-address-per-pool behaviour introduced in `746b552`.

## Introduced in

`694fc45` — v1alpha2: switch non-API code to v1alpha2 types, implement conditions state machine

## Fix

Normalise pool refs to a value-comparable key before using them as map keys, e.g. a struct with dereferenced `APIGroup` string, or a plain `Name/Kind/APIGroup` string tuple.

Issue #778, opened 2026-05-27
bug(ipam): GetIPAMInClusterAPIGroup returns group/version string instead of group


wikkyk:
**What steps did you take and what happened:**

`GetIPAMInClusterAPIGroup()` in `pkg/consts/consts.go` returns a pointer to `ipamicv1.GroupVersion.String()`, which produces the full group/version string `"ipam.cluster.x-k8s.io/v1alpha2"`. This value is used to populate `corev1.TypedLocalObjectReference.APIGroup`, which by the Kubernetes API contract should hold only the group component (`"ipam.cluster.x-k8s.io"`).

The same wrong expression is also inlined directly in two places in `pkg/kubernetes/ipam/ipam.go` (in `GetInClusterPools`).

As a result, any `PoolRef` written by CAPMOX carries a group/version string in the `APIGroup` field rather than a bare group string.

**What did you expect to happen:**

`TypedLocalObjectReference.APIGroup` should contain `"ipam.cluster.x-k8s.io"`, not `"ipam.cluster.x-k8s.io/v1alpha2"`.

**Anything else you would like to add:**

The bug is present in both v0.8 and v0.9 but its observable impact differs by version:

- **v0.8**: `GetIPAddressByPool` filtered on `n.APIVersion` (from TypeMeta, not `Spec.PoolRef.APIGroup`), so the incorrectly written `PoolRef.APIGroup` value was never read for filtering. The write-side bug was effectively dormant.
- **v0.9** (in development): `GetIPAddressByPool` was corrected to read `n.Spec.PoolRef.APIGroup` instead of the cleared TypeMeta field. It uses `strings.Cut` to extract just the group before comparing, which handles both `"group"` and `"group/version"` forms — so filtering works correctly. However, the stored value remains incorrect per the API contract.

**Suggested fix:**

`pkg/consts/consts.go`:
```go
// Before
func GetIPAMInClusterAPIGroup() *string {
	return new(ipamicv1.GroupVersion.String())
}

// After
func GetIPAMInClusterAPIGroup() *string {
	return new(ipamicv1.GroupVersion.Group)
}
```

`pkg/kubernetes/ipam/ipam.go` (two sites in `GetInClusterPools`):
```go
// Before
poolSpec.PoolRef = corev1.TypedLocalObjectReference{
    APIGroup: new(ipamicv1.GroupVersion.String()),
    ...
}

// After
[...]

PR #830, opened 2026-07-03
fix(ipam): store bare API group in PoolRef.APIGroup (#778)

vidarno:
*Issue #, if available:* #778

*Description of changes:*

`GetIPAMInClusterAPIGroup()` returned `ipamicv1.GroupVersion.String()`
(`"ipam.cluster.x-k8s.io/v1alpha2"`), but its value is used to populate
`corev1.TypedLocalObjectReference.APIGroup`. Per the Kubernetes API contract,
`APIGroup` must hold only the group component (`"ipam.cluster.x-k8s.io"`), not
the group/version string.

- `pkg/consts/consts.go`: return `GroupVersion.Group` and clarify the doc comment.
- `pkg/kubernetes/ipam/ipam.go`: route the two inlined `GroupVersion.String()`
  occurrences in `GetInClusterPools` through `GetIPAMInClusterAPIGroup()` so there
  is a single source of truth.

**Scope / impact:** This corrects the value persisted into the `spec.poolRef.APIGroup`
of `IPAddressClaim`/`IPAddress` objects; it is not a behavioural fix. On current `main`
the malformed value is tolerated everywhere it is read: `GetIPAddressByPool` filters on
the IPAddress TypeMeta group, and `matchesPoolRef`/`matchesClaimPoolRef` run both sides
through `normalizedAPIGroup()`, which strips any version component before comparing. The
fix makes the stored data contract-correct at the source, so that tolerance is only
needed for genuinely external/legacy input, and removes a latent trap for any future
exact-string comparison on `APIGroup` (cf. #796).

*Testing performed:*

- Added unit tests in `pkg/consts/consts_test.go` locking the `APIGroup` contract
  (previously the package had no tests).
- `go build ./...`, `go vet ./...`, and `gofmt` clean.
- `pkg/consts`, `pkg/kubernetes/ipam`, and `internal/service/vmservice` unit tests pass.
  (`internal/controller` requires envtest binaries and is unaffected by this change.)


wikkyk:
Technically, you are right. But this is a deeper cut than it would first appear.

Issue #735, opened 2026-04-28
ProxmoxMachineTemplate webhook violates CAPI dry-run contract, blocks all ClusterClass-driven upgrades


pensfan2085:
### Summary

`ProxmoxMachineTemplate.ValidateUpdate` enforces an unconditional `reflect.DeepEqual` immutability check on `Spec`. This denies Cluster API's topology controller dry-run server-side-apply, which is the gate the topology controller uses to decide whether to rotate per-cluster `ProxmoxMachineTemplate`s during a ClusterClass-driven upgrade. With the dry-run denied, rotation never runs, and `Cluster.status.conditions[type=TopologyReconciled].reason` stays at `ReconcileFailed` indefinitely.

The CAPI InfraMachine provider contract requires `InfrastructureMachineTemplate` validation webhooks to skip the immutability check when the request is a topology controller dry-run apply (signaled by the `topology.cluster.x-k8s.io/dry-run` annotation on the object plus a dry-run flag on the request). CAPV, CAPA, and CAPD all implement this. CAPMox does not.

The result: every Kubernetes version (or any other `spec.template.spec.*`) upgrade on every ClusterClass-managed workload cluster wedges and requires manual reconstruction of the per-cluster `ProxmoxMachineTemplate`s.

### Environment

- Cluster API: v1.11.x (v1beta2 contract)
- CAPMox: v0.8.1 — code path also confirmed unchanged against `main` at the time of filing (`internal/webhook/proxmoxmachinetemplate_webhook.go`)
- Cluster topology: ClusterClass-managed workload cluster, KCP + at least one MachineDeployment

wikkyk:
I set this issue's type to bug because we want to have good ClusterClass support. Strictly speaking, it's not a misbehaviour, it's more of a missing functionality that you'd reasonably expect to have.

PR #832, opened 2026-07-03
fix(webhook): skip ProxmoxMachineTemplate immutability on topology dry-run (#735)

vidarno:
*Issue #, if available:* #735

*Description of changes:*

`ProxmoxMachineTemplate.ValidateUpdate` enforced an unconditional
`reflect.DeepEqual` immutability check on `Spec` and ignored the admission
request. This denies Cluster API's topology controller dry-run server-side-apply
— the gate the topology controller uses to decide whether to rotate per-cluster
`ProxmoxMachineTemplate`s during a ClusterClass-driven upgrade. With the dry-run
denied, rotation never runs and `Cluster.status.conditions[type=TopologyReconciled]`
stays at `ReconcileFailed`, wedging every ClusterClass-managed upgrade.

The CAPI InfraMachine provider contract requires template validation webhooks to
skip the immutability check for topology-controller dry-run requests (signalled by
the `topology.cluster.x-k8s.io/dry-run` annotation plus the request dry-run flag).
CAPV, CAPA, and CAPD all implement this.

- Read the admission request from the context (`admission.RequestFromContext`) and
  gate the immutability check with CAPI's own helper `topology.IsDryRunRequest`.
  Regular `kubectl --dry-run` still enforces immutability, since it does not carry
  the topology annotation.
- No new dependency: `sigs.k8s.io/cluster-api/util/topology` is part of the already
  required `sigs.k8s.io/cluster-api` module (`go mod tidy` is a no-op).

*Testing performed:*

- New standalone unit tests in `internal/webhook/proxmoxmachinetemplate_webhook_test.go`
  (no envtest required) covering:
  - no spec change → allowed
  - spec change on a normal update → rejected (immutable)
  - spec change on a plain (non-topology) dry-run → still rejected
  - spec change on a topology-controller dry-run → allowed
  - topology annotation without a dry-run request → rejected
  - missing `admission.Request` in context → `BadRequest`, not a panic
- `go build ./...`, `go vet ./...`, and `gofmt` clean.


vidarno:
The gate this PR adds mirrors how the in-tree CAPI providers implement *MachineTemplate ValidateUpdate: read the admission request from the context, then skip the immutability check when topology.IsDryRunRequest returns true. All three import the helper from sigs.k8s.io/cluster-api/util/topology — the same module CAPMox already depends on, so this adds no new dependency. Links are pinned to a commit so the line numbers stay valid.

CAPD — dockermachinetemplate.go#L58-L74
```
func (webhook *DockerMachineTemplate) ValidateUpdate(ctx context.Context, oldObj *infrav1.DockerMachineTemplate, newObj *infrav1.DockerMachineTemplate) (admission.Warnings, error) {
      req, err := admission.RequestFromContext(ctx)
      if err != nil {
              return nil, apierrors.NewBadRequest(fmt.Sprintf("expected a admission.Request inside context: %v", err))
      }
      var allErrs field.ErrorList
      if !topology.IsDryRunRequest(req, newObj) {
              // ... spec comparison + immutability error
      }
```

CAPV — vspheremachinetemplate.go#L109-L119
```
func (webhook *VSphereMachineTemplate) ValidateUpdate(ctx context.Context, oldTyped, newTyped *infrav1.VSphereMachineTemplate) (admission.Warnings, error) {
      req, err := admission.RequestFromContext(ctx)
      if err != nil {
              return nil, apierrors.NewBadRequest(fmt.Sprintf("expected a admission.Request inside context: %v", err))
      }
      var allErrs field.ErrorList
      if !topology.IsDryRunRequest(req, newTyped) &&
              !reflect.DeepEqual(newTyped.Spec.Template.Spec, oldTyped.Spec.Template.Spec) {
              allErrs = append(allErrs, field.Invalid(field.NewPath("spec", "template", "spec"), newTyped, machineTemplateImmutableMsg))
      }
```

CAPA — awsmachinetemplate_webhook.go#L264-L273
 ```
     req, err := admission.RequestFromContext(ctx)
      if err != nil {
              return nil, apierrors.NewBadRequest(fmt.Sprintf("expected a admission.Request inside context: %v", err))
      }
      ...
      if !topology.IsDryRunRequest(req, newAWSMachineTemplate) && !cmp.Equal(newAWSMachineTemplate.Spec, oldAWSMachineTemplate.Spec) {
```

The helper itself: util/topology/topology.go#L29 — IsDryRunRequest(req, obj) returns true only when the request is a dry-run and the object carries the topology.cluster.x-k8s.io/dry-run annotation, so plain kubectl --dry-run still enforces immutability.

This PR is the identical shape: admission.RequestFromContext(ctx) → topology.IsDryRunRequest(req, newMachineTemplate) gating the existing reflect.DeepEqual, including the same BadRequest on a missing request. We use IsDryRunRequest rather than the older ShouldSkipImmutabilityChecks — the latter has since been removed from util/topology on CAPI main, and all three providers above have migrated to IsDryRunRequest.

wikkyk:
Sorry for the merge commit but github didn't want to let me rebase.

wikkyk:
Please make sure your PR passes `make verify`.

vidarno:
`make verify `now passes

vidarno:
Validated end to end on a real cluster.

**Setup:** CAPI v1.13.3, capmox CRDs at v0.8.2, manager image built from this branch
(`73c9147`) and published at `ghcr.io/vidarno/cluster-api-provider-proxmox:dryrun-735`
(public, linux/amd64). Only the manager image changed - no CRD churn. Workload cluster
is ClusterClass-managed, 3 control plane + 2 workers on Proxmox, IPv6-only.

**Before the patch**, bumping `Cluster.spec.topology.version` from v1.34.0 to v1.35.0
wedged immediately and never recovered:

TopologyReconciled=False ReconcileFailed: error reconciling the Cluster topology:
failed to create patch helper for ProxmoxMachineTemplate capi-lab-topo/lab-topo-6x84x:
server side apply dry-run failed for modified object: admission webhook
"validation.proxmoxmachinetemplate.infrastructure.cluster.x-k8s.io" denied the request:
ProxmoxMachineTemplate.infrastructure.cluster.x-k8s.io "lab-topo-6x84x" is invalid:
spec: Forbidden: ProxmoxMachineTemplate is immutable

The upgrade cannot avoid rotation: the node image is selected by
`templateSelector.matchTags`, and the `vMAJOR.MINOR` tag moves with the Kubernetes
version, so the machine template spec changes on every upgrade.

**After the patch**, with nothing else altered, the rotation went through and the
topology controller ran the upgrade in the documented order - control plane to
completion first, workers only after:

07:36:42Z  CP[v1.35.0 up=1/3 ready=3]  MD[v1.34.0 up=2/2 ready=2]   CP starts
07:39:47Z  CP[v1.35.0 up=2/3 ready=3]  MD[v1.34.0 up=2/2 ready=2]
07:41:52Z  CP[v1.35.0 up=3/3 ready=3]  MD[v1.34.0 up=2/2 ready=2]   CP done
07:43:57Z  CP[v1.35.0 up=3/3 ready=3]  MD[v1.35.0 up=1/2 ready=2]   workers start
07:46:01Z  CP[v1.35.0 up=3/3 ready=3]  MD[v1.35.0 up=2/2 ready=2]
07:48:08Z  all 5 machines Running on v1.35.0

Final state `TopologyReconciled=True ReconcileSucceeded`, and the cluster passes a full
health check afterwards (nodes Ready, Cilium, CoreDNS, pod-to-pod and service
connectivity, network policy enforcement, external DNS/egress).


Is this something that could go into 0.10.0, @wikkyk ? Happy to rebase or adjust the approach if
you would rather gate it differently.

3deep5me:
Would love to see this - any chances to get this into the next release?

wikkyk (approved):

ozhankaraman:
Confirmed we hit exactly this on our end too, TopologyReconciled stuck at ReconcileFailed on a live cluster trying to bump k8sVersion/the machine template ID. #832 looks like the right fix. Any chance this makes it into the next release? 

It's currently blocking all our ClusterClass-driven cluster upgrades, so a release with this fix would unblock us, thank you for all your support guys

PR #834, opened 2026-07-06
feat(api): add PCI device passthrough support for ProxmoxMachine VMs

steve-fraser:
## Summary

This PR is far from merge-ready. I want to gauge the appetite for this 

Adds a `PCIDevices` field to `ProxmoxMachineSpec` allowing host PCI devices to be passed through to VMs via the Proxmox `hostpciN` config API. Supports PCIe mode, ROM BAR control, and primary GPU (`x-vga`) designation.

*Issue #, if available:* N/A

*Description of changes:*

- New `PCIDevice` struct in `api/v1alpha2` with fields: `deviceID`, `pcie`, `rombar`, `primaryGPU`
- New `pciDevices []PCIDevice` field on `ProxmoxMachineSpec`, applied pre-boot alongside CPU/memory/network config
- `formatPCIDevice` helper formats the Proxmox `hostpciN` option string (e.g. `0000:01:00,pcie=1,x-vga=1`)
- v1alpha1 round-trip preserved via the existing annotation mechanism — `PCIDevices` is dropped on conversion to v1alpha1 and restored from the annotation on conversion back
- CRDs regenerated via `make manifests generate`

*Testing performed:*

Motivated by a real use case: an Intel GPU that must be manually attached as a PCI device after every node replacement in a `MachineDeployment` with `OnDelete` rollout strategy. This change automates that step.

Unit tests added for `formatPCIDevice` (table-driven, 8 cases) and `reconcileVirtualMachineConfig` with PCI devices. `make lint verify test` passes clean.

> ⚠️ **Note:** This PR is an early draft to gauge appetite for the feature upstream. It is not yet ready to merge.


api/v1alpha2/proxmoxmachine_types.go
```diff
@@ -174,6 +180,28 @@ type DiskSize struct {
 	SizeGB int32 `json:"sizeGb,omitempty"`
 }
 
+// PCIDevice defines a host PCI device to pass through to the VM.
+type PCIDevice struct {
+	// deviceID is the host PCI device address (e.g. "0000:01:00", "01:00.0").
+	// Multiple devices can be mapped using a semicolon-separated list for SR-IOV.
+	// +kubebuilder:validation:MinLength=1
+	// +required
+	DeviceID string `json:"deviceID,omitempty"`
```
wikkyk: If this can take multiple devices, why not use a `[]string`?

wikkyk:
This is definitely very interesting. It could use an e2e test but we couldn't host one, so there isn't really a good way to test this properly, especially on an ongoing basis. We don't really have a use case for this but I can see it being useful for e.g. building a cluster with GPUs for AI.

steve-fraser:
> This is definitely very interesting. It could use an e2e test but we couldn't host one, so there isn't really a good way to test this properly, especially on an ongoing basis. We don't really have a use case for this but I can see it being useful for e.g. building a cluster with GPUs for AI.

Yeah, I see what you're saying about the e2e not being really feasible. If the NICs on your capmox-e2e runner support SR-IOV, we could pass through a virtual function (not the physical NIC itself, so the runner's own connectivity stays intact) and just verify the resulting VM config on the Proxmox side matches what was requested — that'd give us real end-to-end coverage of the reconciler without needing GPU hardware or touching guest-level networking

api/v1alpha2/proxmoxmachine_types.go
steve-fraser: That's fair feedback. I have implemented it. I initially thought it would be simpler to follow what the hostPCI accepts in its raw form, but I think that's an anti-pattern as far as k8s operators go. So, I modified it accordingly

kaechele:
It would be great if this also supported Resource Mapped devices as well, i.e. the option string would look something like `mapping=nvidia-b200,pcie=1` (where `nvidia-b200` is a user defined string set in the Proxmox Datacenter Resource Mappings) instead of `0000:6d:00.0,pcie=1`. As for schema, mapping / PCI ID are mutually exclusive, so it's either mapping or PCI ID.

PR #861, opened 2026-08-14
refactor(cloudinit): render netplan via structured YAML marshalling

### LINKED ISSUE #780: Replace YAML rendering in netplan with actual yaml.

vidarno:
## Summary

Closes #780.

The netplan network-config was produced by a ~100-line `text/template` that
hand-indented YAML, after which `Render()` parsed the result back with
`yaml.Unmarshal` purely to check it was valid — serialising structured data to
text and re-parsing it to hope it round-trips.

This models the netplan v2 schema as Go structs and marshals them instead. The
output is valid by construction (no post-hoc parse check), special characters are
quoted correctly by the encoder, and device ordering is deterministic (yaml.v3
sorts map keys).

## Changes

- `pkg/cloudinit/netplan.go` (new) — netplan v2 types + `buildNetplanConfig` mapping.
- `pkg/cloudinit/network.go` — drop `networkConfigTpl`; `Render` marshals the model.
- `pkg/cloudinit/types.go` — drop the now-unused `BaseCloudInitData.NetworkConfigData`.
- `pkg/cloudinit/network_test.go` — goldens regenerated to the marshalled output.
- `pkg/cloudinit/netplan_equivalence_test.go` (new) — parses the previous template's
  output and the new goldens into a canonical structure and asserts equality,
  proving only formatting changed, not the netplan config.

## Behaviour

Unchanged. The same `network.Network` validation runs first, `dhcp4`/`dhcp6` are
still emitted explicitly, and two-space indentation is preserved.

The formatting differences, all intended:

- `routes:` / `routing-policy:` entries are block mappings rather than inline
  `{ "to": …, "via": … }` flow style.
- Quoting is the yaml.v3 encoder's rather than hand-written single quotes.
- `ethernets:` is omitted entirely when there are no ethernet devices; the
  template always printed the bare key. Netplan accepts both.
- Device order is yaml.v3's sorted map order rather than input slice order.

## Verification

`netplan_equivalence_test.go` covers this at the unit level, including VRFs with
routes and routing-policy.

It was additionally validated on a live cluster — Proxmox VE 9.2, Debian 13,
dual-NIC IPv6-only: static addresses, an IPv6 default route with a metric,
per-NIC nameservers, and MTU 9000 on the second NIC. The injected network-config
was captured from `inject.go`'s `V(4)` log on an identical machine before and
after the change.

<details>
<summary>Injected network-config, before (template) and after (marshalled)</summary>

Before:

```yaml
network:
  version: 2
  renderer: networkd
  ethernets:
    eth0:
      match:
        macaddress: BC:24:11:52:CA:04
      dhcp4: false
      dhcp6: false
      addresses:
        - '2001:db8:3900:60:200::3/64'
      routes:
        - { "to": "::/0",  "via": "2001:db8:3900:60::1",  "metric": 100, }
      nameservers:
        addresses:
          - '2001:db8:0:503::b'
    eth1:
      match:
        macaddress: BC:24:11:F9:4C:55
      dhcp4: false
      dhcp6: false
      addresses:
        - '2001:db8:e01:ff04::6/64'
      nameservers:
        addresses:
          - '2001:db8:0:503::b'
      mtu: 9000
```

After:

```yaml
network:
  version: 2
  renderer: networkd
  ethernets:
    eth0:
      match:
        macaddress: BC:24:11:A2:57:64
      dhcp4: false
      dhcp6: false
      addresses:
        - 2001:db8:3900:60:200::3/64
      routes:
        - to: ::/0
          via: 2001:db8:3900:60::1
          metric: 100
      nameservers:
        addresses:
          - 2001:db8:0:503::b
    eth1:
      match:
        macaddress: BC:24:11:FD:47:53
      dhcp4: false
      dhcp6: false
      addresses:
        - 2001:db8:e01:ff04::6/64
      nameservers:
        addresses:
          - 2001:db8:0:503::b
      mtu: 9000
```

</details>

With MAC addresses masked, the two parse to identical structures under both
yaml.v3 and PyYAML `safe_load` — the latter being what cloud-init itself uses.
The unquoted IPv6 values still parse as strings (`'::/0'`,
`'2001:db8:3900:60::1'`) while `metric` and `mtu` stay integers, so dropping the
template's hand-quoting does not change any value's type.

On the provisioned node: `cloud-init status --long` reported `done` with no
errors, the expected routes and MTU were applied, and the node joined and went
Ready. Comparing `/etc/netplan/50-cloud-init.yaml` — cloud-init's own re-render,
so formatting-normalised — between a node created before the change and one
created after shows differences only in MAC address and assigned IPv6 address.

## Scope

Ignition's networkd renderer (`pkg/ignition`) is intentionally left as-is; #790
tracks its structured rendering.

## AI assistance

Written with AI assistance (Claude), attributed with a `Co-Authored-By:` trailer
on the commit. I reviewed the change and ran the validation described above.


pkg/cloudinit/netplan.go
```diff
@@ -0,0 +1,182 @@
...
+package cloudinit
+
+import (
+	"github.com/ionos-cloud/cluster-api-provider-proxmox/pkg/network"
+)
+
+// The types below model the subset of the netplan v2 schema that CAPMOX emits.
+// They are marshalled to YAML directly (see NetworkConfig.Render) instead of
+// being rendered through a text template, so the output is valid by
+// construction and special characters are quoted correctly.
```
wikkyk:
```suggestion
// They are marshalled to YAML directly (see NetworkConfig.Render),
// so the output is valid by construction and special characters are quoted correctly.
```

pkg/cloudinit/netplan.go
```diff
@@ -0,0 +1,182 @@
...
+//
+// Field order within each struct matches the desired YAML key order, and
+// pointer / slice fields use `omitempty` so that a key is emitted only when the
+// corresponding ConfigData value is set — mirroring the previous template's
+// conditionals.
```
wikkyk:
```suggestion
// corresponding ConfigData value is set.
```

pkg/cloudinit/netplan.go
```diff
@@ -0,0 +1,182 @@
...
+	Ethernets map[string]netplanEthernet `yaml:"ethernets,omitempty"`
+	VRFs      map[string]netplanVRF      `yaml:"vrfs,omitempty"`
+}
+
+type netplanEthernet struct {
+	Match *netplanMatch `yaml:"match,omitempty"`
+	// dhcp4/dhcp6 are always emitted (netplan treats an absent key as false, but
+	// the previous renderer emitted both explicitly, so we keep that behaviour).
+	DHCP4         bool                `yaml:"dhcp4"`
+	DHCP6         bool                `yaml:"dhcp6"`
```
wikkyk, line 47: This part I'm not sure about. @65278 what do you think?

wikkyk (commented):
I love this. I just need @65278 to take a second look.

pkg/cloudinit/network_test.go
```diff
@@ -37,14 +37,15 @@ const (
       dhcp4: false
       dhcp6: false
       addresses:
-        - '10.10.10.12/24'
+        - 10.10.10.12/24
```
wikkyk, line 40: We've had problems with netplan's yaml reader getting confused by the lack of quoting. Can you have the marshaller always quote IP addresses?

vidarno: The encoder already quotes values that would otherwise be misread. I checked the yaml.v3 output on this branch and parsed it with PyYAML safe_load, the parser cloud-init uses:

## Quoted by the encoder: these would be invalid or change type if left bare
v: '2a01:7e01:e003:3b04::'      # the #79 case: bare, it's a ScannerError
v: '::'
v: "12:34:56:12:34:56"          # bare, PyYAML reads the int 9783981296 (YAML 1.1 base-60)
v: "1:2:3:4:5:6:7:8"            # bare, PyYAML reads the int 2895057742028
v: "2001:0:0:0:0:0:0:1"
v: "010"
v: "0x10"
v: "1e3"

## Left bare: unambiguous, and PyYAML reads each as a string
v: 10.10.10.12/24
v: 8.8.8.8
v: 2001:db8::1
v: ::/0
v: ::1
v: fd00::1

Every value above comes back from PyYAML as the original string. As far as I can tell, the failures we had came from the old template pasting values in without escaping (#79), not from a YAML parser misreading well-formed input. I'm still happy to force quoting on all address fields: it doesn't change any parsed value, and it brings the goldens back closer to main. If you have the specific input that broke netplan for you, I'll add it as a regression test.

PR #871, opened 2026-08-29
ci: build container images for amd64 and arm64

QuentinBtd:
## Summary

Add native container image builds for:

- `linux/amd64`
- `linux/arm64`

The published image reference remains unchanged:

```console
docker pull ghcr.io/ionos-cloud/cluster-api-provider-proxmox:<tag>
```

Docker and containerd will automatically select the appropriate architecture.

## Background

This PR follows up on #491.

The previous implementation explored QEMU-based multi-platform builds and encountered issues with emulation, Buildx `load`, containerd, and per-architecture image scanning.

Following the latest maintainer feedback in #491, this implementation uses GitHub-hosted native ARM64 runners.

## Implementation

- Build AMD64 on `ubuntu-24.04`.
- Build ARM64 on `ubuntu-24.04-arm`.
- Test and scan each image on its native architecture.
- Push each platform image by digest.
- Create a single multi-platform manifest using the existing release tags.
- Do not push images from pull request workflows.
- Do not use QEMU or cross-architecture emulation.

## Validation

- Native ARM64 image build and runtime smoke test
- Trivy scan for HIGH and CRITICAL vulnerabilities
- Unit tests and linters
- Actionlint and Yamllint validation
- Verification that Kubernetes manifests keep using the standard image reference

The final GHCR manifest publication can only be fully validated by a release tag workflow.

## AI assistance

OpenCode, powered by `openai/gpt-5.6-sol`, assisted with the implementation and drafting of this pull request.

Additional AI models were used to review the changes. They did not directly modify the code.

Changes suggested by AI:
- Pin GitHub Actions to full commit SHAs instead of mutable version tags.
- Rename the `REPO` variable to `IMAGE` to make its purpose clearer.
- Help implement the manifest job.
- Suggest shorter and clearer step names, as my initial names tended to be overly verbose.

The final changes were reviewed and validated by the contributor.

### DIFF
--- a/.github/workflows/container-image.yaml
+++ b/.github/workflows/container-image.yaml
@@ -13,20 +13,33 @@ permissions:
 
 jobs:
   image:
-    name: build image
-    runs-on: "ubuntu-latest"
+    name: build image (${{ matrix.arch }})
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - arch: amd64
+            platform: linux/amd64
+            runner: ubuntu-24.04
+          - arch: arm64
+            platform: linux/arm64
+            runner: ubuntu-24.04-arm
+    runs-on: ${{ matrix.runner }}
     steps:
       - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
 
       - name: Generate lowercase repository name
         run: |
-          echo "REPO=${GITHUB_REPOSITORY,,}" >> "${GITHUB_ENV}"
+          echo "IMAGE=ghcr.io/${GITHUB_REPOSITORY,,}" >> "${GITHUB_ENV}"
 
       - name: Docker meta
         id: meta
         uses: docker/metadata-action@dc802804100637a589fabce1cb79ff13a1411302 # v6.2.0
         with:
-          images: ghcr.io/${{ env.REPO }}
+          images: ${{ env.IMAGE }}
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@37fe631027851001ddb9b187196cc803df7f5f0e # v4.3.0
 
       - name: Login to GitHub Container Registry
         if: github.event_name != 'pull_request'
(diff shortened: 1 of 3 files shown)


.github/workflows/container-image.yaml
```diff
@@ -36,11 +49,113 @@ jobs:
...
+      - name: Test image
+        run: docker run --rm "${{ steps.image-ref.outputs.ref }}" --help
+
+      - name: Scan image
+        uses: aquasecurity/trivy-action@ed142fd0673e97e23eac54620cfb913e5ce36c25 # v0.36.0
+        with:
+          image-ref: ${{ steps.image-ref.outputs.ref }}
+          exit-code: '1'
+          ignore-unfixed: true
+          severity: 'HIGH,CRITICAL'
```
wikkyk: Definitely not.

go.mod
wikkyk, line 1: Why are there go.mod changes?

wikkyk (changes requested):
Good idea but this PR adds significant complexity to the workflow. I'm not convinced that this is the most efficient way to do this.

.github/workflows/container-image.yaml
QuentinBtd: Miss understanding [of this comment ](https://github.com/ionos-cloud/cluster-api-provider-proxmox/pull/491#issuecomment-3035623555).

So, you don't want Trivy scan?

go.mod
QuentinBtd: The go.mod change fixes a vulnerability detected by Trivy. However, it is unrelated to multi-architecture support and should indeed be handled in a separate PR.

QuentinBtd:
> Good idea but this PR adds significant complexity to the workflow. I'm not convinced that this is the most efficient way to do this.

Some tools, such as Ko, can generate multi-architecture images. However, you previously suggested using native ARM64 runners in [this comment](https://github.com/ionos-cloud/cluster-api-provider-proxmox/pull/491#issuecomment-4851692836):

> Compiling in qemu is completely out of question.
However, instead of cross-compiling, we could use the new ubuntu-*-arm runners to compile and test natively on arm64.

This implementation follows that recommendation by building and testing each architecture on its corresponding native GitHub-hosted runner. 😄

.github/workflows/container-image.yaml
wikkyk: No, we removed Trivy after their numerous security incidents.

pborn-ionos:
The most efficient way surely would be to just pass `platforms: linux/amd64,linux/arm64` to `docker/build-push-action` and pairing it with an instance of `docker/setup-buildx-action`, no? :)

PR #886, opened 2026-09-10
fix: never destroy a VM that carries another machine's name

### LINKED ISSUE #856: Failed-clone ProxmoxMachine retains stale spec.virtualMachineID; nextid reuse can cause deletion of a different machine's VM
**Environment**

- CAPMOX v0.9.x
- Proxmox VE 9 (trixie)
- Talos-based workload clusters (as applicable)

**Summary**

When a `ProxmoxMachine`'s initial clone fails, the object retains `spec.virtualMachineID` set to the VMID that was reserved for (but never successfully used by) that machine. Because that VMID was never actually consumed on the Proxmox side (no VM config exists there), Proxmox's `nextid` allocator can later hand the same VMID to a different, unrelated `ProxmoxMachine`, which clones and provisions successfully at that VMID.

If the original (failed) `ProxmoxMachine` is later deleted — via CAPI garbage collection, a retry-then-recreate flow, or manual cleanup — the deletion reconciler uses the stale `spec.virtualMachineID` to locate "its" VM. That VMID now belongs to the second, unrelated machine. Depending on whether the delete path verifies the VM's name/owner before issuing the destroy call, this can result in deleting the wrong machine's live VM.

**Repro steps**

1. Create a `ProxmoxMachine` whose clone is guaranteed to fail (e.g. point it at a non-existent template image, or a node/storage combination with no capacity). Confirm `spec.virtualMachineID` is populated and the clone fails, leaving the machine in an error/failed state with no real VM at that ID on Proxmox.
2. Create a second, unrelated `ProxmoxMachine` (different cluster/machine deployment) under conditions where Proxmox's `nextid` allocates the same VMID freed up by the failed clone in step 1 (easiest to reproduce in a fresh or low-VMID-usage range). Confirm it clones and provisions successfully at that VMID.
3. Delete the first (failed) `ProxmoxMachine`.
4. Observe whether the controller's delete path destroys the VM now running at that VMID — which belongs to the second machine — instead of detecting the mismatch and skipping/erroring.

**Suggested fix**

- On clone failure, clear `spec.virtualMachineID` (or otherwise mark it explicitly unallocated) so the stale value can't be used as a delete target later.
- Before issuing any destroy call, verify the live VM's name/owner matches the `ProxmoxMachine` being deleted, and skip/no-op instead of destroying on a mismatch. (A similar name-verification guard already exists on the provisioning path, per #842 — extending the same check to the deletion path would close this gap.)

**Related**

#842 documents a different but related VMID hazard: a TOCTOU race in *concurrent* next-free-VMID selection, where two machines are allocated the same VMID at creation time and the loser gets permanently wedged in an error state (never reaching deletion). This report is about a *sequential* hazard instead — a stale pin surviving a failed clone, being recycled by `nextid` for an unrelated machine, and then being used unsafely at delete time. The fixes are complementary: #842's suggested "release/re-select on collision" and this report's "clear the pin on failure, verify identity before delete" both reduce to the same underlying gap — VMID pinning and deletion both need to reason about whether the pin still corresponds to a real, owned VM.

oivindoh:
Fixes #856

`DeleteVM` sent the recorded `spec.virtualMachineID` straight to Proxmox. Proxmox reuses a freed VMID for the next clone within seconds, and a MachineHealthCheck replaces a machine that fast, so a record can go stale while a healthy VM of another machine runs under the ID. The destroy then kills that VM, and since the victim fails its own health check next, one stale record becomes a chain across clusters (task log in #856).

`FindVM` already refuses a VM with another name on the create path. This applies the same check on delete: when the VM at the recorded ID has another name, drop the node location and the finalizer and never call destroy. Two regression tests, one for the mismatch and one for the plain destroy; the existing not-found test gains the `GetVM` expectation.


oivindoh:
@wikkyk thanks for taking a look; doesn't seem the e2e wants to actually start though

internal/service/vmservice/delete.go
```diff
@@ -36,13 +36,17 @@ func DeleteVM(ctx context.Context, machineScope *scope.MachineScope) error {
 	vmID := machineScope.ProxmoxMachine.GetVirtualMachineID()
 	node := machineScope.LocateProxmoxNode()
 
+	// Proxmox reuses a freed VMID for the next clone within seconds, so the
+	// recorded ID may now belong to another machine. Never destroy a VM whose
+	// name is not ours; treat it as already gone.
+	if vm, err := machineScope.InfraCluster.ProxmoxClient.GetVM(ctx, node, vmID); err == nil && vm.Name != machineScope.ProxmoxMachine.GetName() {
+		machineScope.Info("VMID belongs to another VM, skipping destroy", "vmID", vmID, "vmName", vm.Name)
+		return releaseDeletedVM(machineScope)
+	}
```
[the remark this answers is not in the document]
oivindoh: Good catch. Fixed.

DeleteVM no longer trusts a failed lookup. On a GetVM error it now asks the cluster: CheckID reports whether the ID is free, and FindVMResource names the holder. Only a name match reaches the destroy call. An unreadable cluster returns an error and requeues instead of deleting, so the path is fail-closed.

This also covers a case the comment did not name: a replacement VM cloned onto a different node, where GetVM on the recorded node fails while the ID is taken.

I did not pass the expected name into the client DeleteVM. That changes the exported Client interface and regenerates the mock, which is wider than this fix needs, and it still would not make stop-then-delete atomic.

internal/service/vmservice/delete.go
```diff
@@ -31,18 +31,30 @@ import (
...
+// minimumVMID is the lowest VMID Proxmox accepts.
+const minimumVMID = 100
+
 // DeleteVM implements the logic of destroying a VM.
 func DeleteVM(ctx context.Context, machineScope *scope.MachineScope) error {
 	vmID := machineScope.ProxmoxMachine.GetVirtualMachineID()
 	node := machineScope.LocateProxmoxNode()
 
+	// GetVirtualMachineID returns -1 for a machine that never got an ID.
+	if vmID < minimumVMID {
```
wikkyk: Then I would prefer this to check for `-1`.

oivindoh: Yes - check is now for -1 👍

Issue #370, opened 2024-12-18
[Proposal] Support for multi Proxmox Clusters (Datacenters)


mcbenjemaa:
**Details**
[A clear and concise description of what you want to happen.]

This issue tracks the creation of the proposal of supporting multi Proxmox Clusters.

Ref: #231

wikkyk:
I updated the original content with the proposal.

MariusELM:
Hi,

We're very interested in this feature internally, as it aligns well with our goals for better resource utilization and high availability across Proxmox clusters.

We're currently working on implementing support for multi-instance provisioning in CAPMOX, and we've made substantial progress. Specifically, we've nearly completed development of the v1alpha2 version of the ProxmoxCluster and ProxmoxMachine CRDs with MultiInstance mode support.

At this stage:

-   Core functionality for multi-instance provisioning is mostly in place.
    
-   The APIClient now supports multiple Proxmox clusters via a map-based client implementation.
    
-   The CRDs have been extended to include the necessary fields (spec.settings, instances, etc.).


However, we have not yet completed the unit tests and end-to-end (e2e) testing. These will be the next areas of focus to ensure stability and correctness across diverse scenarios.

Since this issue currently stands as a proposal, we would like to know the maintainers’ thoughts on:

-   The proposed CRD changes and versioning strategy (e.g., v1alpha2 vs. v1alpha3)
    
-   The best way to structure and submit a pull request for review
    
-   Any expectations for backwards compatibility or migration support
    

We’re happy to adapt our work to meet project standards and would appreciate any guidance on the preferred path forward. Looking forward to collaborating on this!

wikkyk:
@MariusELM Right now we're finishing [v1alpha2](https://github.com/ionos-cloud/cluster-api-provider-proxmox/milestone/7) which focuses on generalising networking.

We've had some discussions internally about what we would like to see in v1alpha3 and when we would like to see it. I don't think we'll be able to fit multidc in v1alpha2 simply because it's already a fairly extensive change. However, I can see multidc being a v1alpha3 feature. I don't want to commit to any timelines but we would like to start working on v1alpha2 right after releasing v1alpha2.

Please continue your work in the meantime with the assumption that it'll go into v1alpha3. Feel free to look at the v1alpha2 branch but do keep in mind it's still heavily WIP and a lot of functionality is incomplete.

[...]

PR #887, opened 2026-09-10
Multi-AZ / FailureDomain support

### LINKED ISSUE #717: feat: implement CAPI failure domain contract for v1alpha2 zones
## Problem                                                

  CAPMOX v0.8.0 introduced zones (`zoneConfig`) with per-zone IPAM pools, DNS, and network configuration. However, the CAPI failure domain contract is not implemented:                                                                                                                           
   
  - `ProxmoxCluster.status.failureDomains` is not populated                                                                                                                                                                                                                                       
  - `ProxmoxMachine.spec.failureDomain` field does not exist (InfraMachine contract gap)
  - KubeadmControlPlane cannot automatically distribute control plane nodes across zones                                                                                                                                                                                                          
  - MachineDeployments cannot target a specific zone via `spec.failureDomain`                                                                                                                                                                                                                     
                                                                                                                                                                                                                                                                                                  
  This means that even with multiple zones configured, operators must manually manage node placement and have no automatic HA distribution of control plane replicas.                                                                                                                             
                                                                                                                                                                                                                                                                                                  
  ## Proposal                                                                                                                                                                                                                                                                                     
                                                            
  Implement the minimal CAPI failure domain contract by mapping existing v1alpha2 zones to CAPI failure domains. This stays within a single Proxmox cluster — the multi-DC scenario in #370 is a separate, larger effort for v1alpha3.                                                            
   
  ### API changes (additive, backwards-compatible)                                                                                                                                                                                                                                                
                                                            
  | Field | Type | Description |                                                                                                                                                                                                                                                                  
  |-------|------|-------------|                            
  | `ZoneConfigSpec.nodes` | `[]string` | Proxmox nodes belonging to this zone. VMs assigned to this failure domain are scheduled only on these nodes. |
  | `ZoneConfigSpec.controlPlane` | `*bool` | Whether this zone is eligible for control plane machines. Defaults to `true`. |                                                                                                                                                                     
  | `ProxmoxClusterStatus.failureDomains` | `[]clusterv1.FailureDomain` | CAPI v1beta2 contract field, populated from `zoneConfig`. |                                                                                                                                                             
  | `ProxmoxMachineSpec.failureDomain` | `string` | CAPI InfrastructureMachine contract field. |                                                                                                                                                                                
                                                                                                                                                                                                                                                                                                  
  ### Controller changes                                                                                                                                                                                                                                                                          
                                                                                                                                                                                                                                                                                                  
  - `reconcileFailureDomains` in cluster controller — populates `status.failureDomains` from `zoneConfig` (sorted by name to prevent reconcile loops)                                                                                                                                             
  - Machine controller reads `Machine.Spec.FailureDomain` and scopes `AllowedNodes` + IPAM pool to the zone
  - Uses in-memory `effectiveZone` / `effectiveAllowedNodes` on MachineScope — no spec mutation from the controller                                                                                                                                                                               
  - Retryable `FailureDomainNotReady` condition for zone lookup failures (not terminal, requeues with backoff)                                                                                                                                                                                    
  - Full conversion webhook support for v1alpha1 round-trip                                                                                                                                                                                                                                       
                                                                                                                                                                                                                                                                                                  
  ### Example                                                                                                                                                                                                                                                                                     
                                                                                                                                                                                                                                                                                                  
```yaml
spec:
    zoneConfig:
      - zone: "rack-a"
        nodes: ["pve1", "pve2"]
        ipv4Config:                                                                                                                                                                                                                                                                               
          addresses: ["10.0.1.10-10.0.1.30"]
          prefix: 24                                                                                                                                                                                                                                                                              
          gateway: "10.0.1.1"                               
        dnsServers: ["8.8.8.8"]                                                                                                                                                                                                                                                                   
      - zone: "rack-b"                                                                                                                                                                                                                                                                            
        nodes: ["pve3", "pve4"]
        controlPlane: false                                                                                                                                                                                                                                                                       
        ipv4Config:                                                                                                                                                                                                                                                                               
          addresses: ["10.0.2.10-10.0.2.30"]
          prefix: 24                                                                                                                                                                                                                                                                              
          gateway: "10.0.2.1"                               
        dnsServers: ["8.8.8.8"]
```
                                                                                                                                                                                                                                                                                       
  KCP with 3 replicas distributes automatically across zones with controlPlane: true. Worker MachineDeployments with failureDomain: "rack-b" are placed on pve3/pve4 with IPs from the rack-b pool.                                                                                               
                                                                                                                                                                                                                                                                                                  
  What this enables                                                                                                                                                                                                                                                                               
                                                            
  - KubeadmControlPlane round-robin distribution of CP replicas across zones                                                                                                                                                                                                                      
  - Per-zone worker MachineDeployments via spec.failureDomain
  - Per-zone IPAM pool selection (already implemented, now wired to the failure domain contract)                                                                                                                                                                                                  
  - Worker-only zones via controlPlane: false                                                                                                                                                                                                                                                     
  - Full backwards compatibility — clusters without zoneConfig work exactly as before                                                                                                                                                                                                             
                                                                                                                                                                                                                                                                                                  
  Relationship to #370                                                                                                                                                                                                                                                                            
                                                                                                                                                                                                                                                                                                  
  This is the single-cluster foundation for the multi-DC effort. The multi-DC work in #370 (targeted for v1alpha3) would extend this with per-zone credentials and API endpoints via new CRDs, similar to CAPV's VSphereDeploymentZone / VSphereFailureDomain pattern. This implementation        
  establishes the contract surface that the multi-DC work builds on.
                                                                                                                                                                                                                                                                                                  
  Implementation                                            

  We have a working implementation with unit tests and documentation, validated on a real Proxmox cluster (10 nodes, 2 zones) with 3 CP + 2 workers. KCP correctly distributes CP replicas round-robin across failure domains, workers are pinned to their target zone, and IPAM pools are selected per-zone.
                                                                                                                                                                                                                                                                                                  
  Branch: https://github.com/ElmecOSS/cluster-api-provider-proxmox/tree/feat/failure-domains                                                                                                                                                                                                      
  
  Single commit on top of v0.8.0. Happy to open a PR against upstream if there's interest.                                                                                                                                                                                 
                                                            
  Environment                                                                                                                                                                                                                                                                                     
                                                            
  - CAPMOX: v0.8.0 (v1alpha2)
  - CAPI: v1.11.7+ (v1beta2)
  - Proxmox VE: 9.1.4

robinsft:
## feat: multi-AZ / failure domain support for multi-cluster Proxmox VE

Resolves #717.

### Summary

This PR implements the Cluster API failure domain contract for CAPMOX and, building on it, adds the ability for a single CAPI `Cluster` to span multiple physically separate Proxmox VE clusters — one per availability zone — each with its own API endpoint and credentials.

`ProxmoxCluster.spec.availabilityZones` groups Proxmox nodes into named zones. The cluster controller reconciles these into `ProxmoxCluster.status.failureDomains`, so CAPI can spread control-plane and worker machines across zones (`Machine.spec.failureDomain`). CAPMOX restricts scheduling of a machine to the nodes of its assigned zone. Each zone may optionally declare its own `credentialsRef`, so that a zone is backed by an independent Proxmox VE cluster reached through its own client. Zones without a `credentialsRef` keep using the `ProxmoxCluster`'s default client (global credentials or `spec.credentialsRef`), so the single-cluster path is unchanged.

### Changes

**API (additive, backwards-compatible):**
- `ProxmoxCluster.spec.availabilityZones` — new top-level field (`[]AvailabilityZoneSpec`), a map keyed by `name`, with `nodes` and an optional `credentialsRef`.
- `AvailabilityZoneSpec` — `name`, `nodes` (required, min 1), `credentialsRef` (optional `SecretReference`).
- `ProxmoxCluster.status.failureDomains` (`[]clusterv1.FailureDomain`) — reconciled from `spec.availabilityZones`, never set by the user.
- `ProxmoxMachine.status.failureDomain` (`string`, validated `MinLength=1`/`MaxLength=256`) — recorded from the zone of the node the machine is scheduled on.
- v1alpha1 conversion: `availabilityZones`, `failureDomains` and `failureDomain` are restored across the hub→spoke→hub round-trip (they do not exist in v1alpha1).

**Cluster controller:**
- `reconcileFailureDomains` syncs `status.failureDomains` from `spec.availabilityZones` (all zones advertised as control-plane-eligible).
- `ClusterScope.NewClusterScope` now tolerates the absence of a default client when every `allowedNodes` entry is covered by a zone that has its own `credentialsRef` (`defaultProxmoxClientNeeded`); it fails fast with a clear condition otherwise.

**Proxmox client resolution (core of the multi-cluster support):**
- `ClusterScope.GetProxmoxClient(ctx, zone)` returns a zone-specific client built (and cached) from the zone's `credentialsRef`, falling back to the default client.
- `ClusterScope.GetProxmoxClientForNode(ctx, node)` resolves the client from the zone the node belongs to — used once a machine has been scheduled, and for plain `MachineDeployment` workers which CAPI does not assign a `failureDomain` to.
- `MachineScope.ProxmoxClient(ctx)` and `ProxmoxClientForTarget(ctx, node)` select the correct client (node → failure domain → default) and replace all direct uses of `InfraCluster.ProxmoxClient` across the VM lifecycle: find, create/clone, configure, delete, power state, IP tagging, and task tracking.

**Scheduler:**
- When `Machine.spec.failureDomain` is set, candidate nodes are intersected with the zone's nodes.
- Round-robin balancing is performed per availability zone (not per node) via a `nodeToZone` map, so a multi-node zone is deprioritized as a group during rolling upgrades.
- Memory queries are issued per node through a `multiZoneClient`, so each candidate node is queried against the Proxmox VE cluster that actually owns it.

**Docs and templates:**
- `docs/advanced-setups.md` documents both the single-cluster AZ case and the multi-cluster (per-zone credentials) case.
- The default and Cilium ClusterClasses expose an `availabilityZones` variable patched into the `ProxmoxClusterTemplate`; the example clusters set it.

**Tests:** unit tests for the API helpers, the cluster scope (default-client requirement, per-zone client resolution), the machine scope (client selection by node/failure domain), and the scheduler (per-AZ balancing, failure-domain node restriction).

### Relationship to #720

This PR covers the same CAPI contract surface as @emanuelebosetti's #720 (`status.failureDomains`, the machine failure-domain field, and `reconcileFailureDomains`), and we are grateful to @emanuelebosetti for the original issue, the working implementation, and the research that established the contract mapping and the review feedback it collected. We reuse that foundation but take a different design direction on several points:

| Aspect | #720 | This PR |
|---|---|---|
| API surface | Reuses `zoneConfig`, adding `nodes` and `controlPlane` to `ZoneConfigSpec` | Introduces a separate `spec.availabilityZones` (`AvailabilityZoneSpec`), keeping node grouping/scheduling concerns apart from `zoneConfig`'s IP/network config |
| Multi-DC (#370) | Explicitly out of scope, deferred to v1alpha3 | Delivered in v1alpha2 through per-zone `credentialsRef` |
| Machine failure-domain field | `*string` on `ProxmoxMachineSpec` | `string` on `ProxmoxMachineStatus` with length validation, set by the controller from the scheduled node's zone |
| Zone override pattern | `effectiveZone`/`effectiveAllowedNodes` resolved once at `MachineScope` construction, exposed via `AllowedNodes()`/`Zone()` | No spec mutation; the client is resolved per call site, threaded through the whole VM lifecycle so each operation reaches the correct Proxmox cluster |
| Control-plane eligibility | `ControlPlaneEligible` per zone | Not introduced; all zones are advertised as control-plane eligible |
| `FailureDomainNotReady` condition | Retryable condition with 30s requeue | Not required: zone/client resolution is performed at the call site and returns a typed error |
| Scheduler balancing | Per node | Per availability zone, with per-node client for resource queries |

The net difference is that this PR delivers the multi-cluster scenario (#370) inside v1alpha2, at the cost of threading per-zone client resolution through every Proxmox interaction rather than only the scheduler.

### Backwards compatibility

Clusters without `availabilityZones` behave exactly as before. Zones without a `credentialsRef` fall back to the `ProxmoxCluster` default client. No feature gate, webhook change, or controller restart is required to enable availability zones.

### Testing performed

- `make build` succeeds.
- `make test` passes, including the new unit tests added by this branch:
  - `TestSetInClusterIPPoolRefMultipleZones` — multi-zone `InClusterZoneRef` population.
  - `TestSelectNodeBalancesAcrossAvailabilityZones` — scheduler round-robin at the zone level rather than the node level (two-node AZ is deprioritized as a group).
  - `TestScheduleVM_MultiZoneWithoutFailureDomain` — a worker `Machine` with no `spec.failureDomain` and `allowedNodes` spanning two AZs: the node in the AZ with its own `credentialsRef` is queried through its zone client, and scheduling fails rather than silently using the default client against a node it does not manage.
  - `TestNewClusterScope_NoDefaultCredentialsNeededWhenZonesCoverAllNodes` — `NewClusterScope` does not fail when every `allowedNodes` entry is covered by a zone with its own `credentialsRef`, and requesting the default client explicitly returns a clear error.
  - `TestClusterScope_GetProxmoxClient` / `TestClusterScope_GetProxmoxClientForNode` — client resolution by zone and by node: empty/unknown zone and zone without `credentialsRef` fall back to the default client; zone with `credentialsRef` builds a dedicated client.
  - `TestMachineScope_ProxmoxClient_NominalSingleCluster` / `_PerZoneOverride` / `_ResolvesFromKnownNodeWithoutFailureDomain` — `MachineScope.ProxmoxClient` selects the right client from the scheduled node (preferred), then the failure domain, then the default; nominal single-cluster path unchanged.
  - v1alpha1 fuzz conversion (`hubProxmoxClusterSpec`, `hubProxmoxClusterStatus`, `hubProxmoxClusterTemplateSpec`) updated to normalize `AvailabilityZones` and `FailureDomains` for round-trip stability.
- `make manifests generate` run; CRDs, RBAC, deepcopy, and conversion outputs regenerated.
- E2E validation on a real multi-node Proxmox VE setup (not run in CI here): control-plane machines spread across failure domains, worker `MachineDeployment` machines scheduled within their assigned zone, per-zone IPAM pools selected correctly, and the single-cluster path (no `availabilityZones`) provisioned unchanged.


emanuelebosetti:
Author of #720 here, adding some context on the multi-cluster part.

Since #720 we (Elmec) have been running a soft fork that builds the same multi-cluster feature on top of #720; we are currently testing it on our own Proxmox clusters. It has not been submitted upstream yet, but it may be useful as a reference while this PR is reviewed, because it hit several of the same design points and made different calls:

- Branch: https://github.com/ElmecOSS/cluster-api-provider-proxmox/tree/elmec-main (based on upstream `main` as of mid-July, rebase pending; the three `ci(elmec)` workflow commits are fork-specific and can be ignored)
- Full diff vs upstream `main`: https://github.com/ionos-cloud/cluster-api-provider-proxmox/compare/main...ElmecOSS:cluster-api-provider-proxmox:elmec-main
- Design/usage doc: https://github.com/ElmecOSS/cluster-api-provider-proxmox/blob/elmec-main/docs/multi-dc.md

The main differences, as far as I can tell from reading this PR:

| Topic | This PR | elmec-main |
|---|---|---|
| API | new `spec.availabilityZones` next to `zoneConfig` | extends `zoneConfig` (`nodes`, `controlPlaneEligible`, `credentialsRef`, `templateSource`) so placement, IPAM pool and credentials share one zone identity |
| Client resolution | per call site; unknown zone or node outside any AZ falls back to the default client | resolved once per MachineScope; unknown zone or unresolvable secret is an error (`ZoneClientUnavailable`), never another endpoint |
| Deletion | uses the resolved client, with the fallback above | finalizer is kept until the zone client resolves; the webhook refuses removing a zone or changing its `credentialsRef` while `status.nodeLocations` still records machines in it |
| Client lifecycle | cache lives in `ClusterScope`, i.e. one new client (and `Version` probe) per reconcile; default transport | manager-wide caching factory keyed on secret data, negative cache, dial/TLS/header timeouts, eviction |
| Zone health | – | standalone `ZonesAvailable` condition, periodic probe, events; deliberately not part of `Ready` |
| Zone secrets | no finalizer/ownerRef on zone secrets | all referenced secrets adopted and released when the ref changes, cross-namespace handled |
| Templates | `sourceNode`/`templateID` stay per machine, so separate PVE clusters need `templateSelector` | per-zone `templateSource` override (templates cannot be cloned across PVE clusters) |
| IPAM / DNS | pool still selected from `spec.network.zone` | pool and `dnsServers` follow the failure domain |
| Scheduler | per-AZ balancing, per-node client (nice, we don't have this) | allowedNodes restricted to the zone, per-node balancing as upstream |

Two things I would flag on this PR based on what we ran into while testing:

1. **Fail-open deletion.** If an AZ is removed from the spec while it still has machines, or a machine lands on a node not listed in any AZ, every operation including `DeleteVM` goes to the default client. On the wrong PVE cluster a `GetVM` returning "not found" drops the finalizer and orphans a running VM, and a VMID collision can delete an unrelated VM. That is why we went fail-closed plus a webhook guard.
2. **IPAM.** KCP spreads control-plane machines across `status.failureDomains`, but pool selection still keys on `spec.network.zone`, which a single KCP template cannot vary per replica, so all control-plane machines draw from the same pool. Per-zone pools only work with separate MachineDeployments.

Also a small doc/code mismatch: `advanced-setups.md` says an unresolvable zone secret falls back to the default client, but `GetProxmoxClient` returns an error in that case.

On the other hand, `status.failureDomain` on the ProxmoxMachine and the per-AZ balancing are things this PR gets right and our branch does not have yet.

Happy to help converge: whichever API shape the maintainers prefer, the fail-closed resolution, the client factory, the webhook guards and the secret lifecycle from elmec-main could be rebased onto it as follow-ups.

wikkyk:
This is a huge feature, so I won't just slap an approve on it and merge it.

This PR and #720 are incompatible. I get it that they're scratching slightly different itches but they're both roughly in the same spot. @robinsft @emanuelebosetti @MariusELM I need you to embrace the power of open source :) and work out a solution that will satisfy both of your use cases. 

I started #896 as a place for that discussion. I need everyone interested in multi-DC/availability zone/failure domain support to participate. I will try to invite everyone that I can find who's expressed interest.

PR #888, opened 2026-09-10
feat(failure-domain): map availability zones to CAPI failure domains

### LINKED ISSUE #717: feat: implement CAPI failure domain contract for v1alpha2 zones

robinsft:
## feat: map availability zones to CAPI failure domains

Resolves #717 (single-cluster scope).

### Summary

This PR implements the Cluster API failure domain contract for CAPMOX within a single Proxmox VE cluster. `ProxmoxCluster.spec.availabilityZones` groups Proxmox nodes into named zones; the cluster controller reconciles these into `ProxmoxCluster.status.failureDomains` so Cluster API can spread control-plane and worker machines across zones via `Machine.spec.failureDomain`. CAPMOX restricts scheduling of a machine to the nodes of its assigned zone and balances round-robin per zone. FindVM records the scheduled node's zone on `ProxmoxMachine.status.failureDomain`.

All zones are reached through the ProxmoxCluster's default client (global credentials or `spec.credentialsRef`); this PR does not change how the Proxmox API is contacted. Per-zone credentials, letting a single CAPI `Cluster` span multiple physically separate Proxmox VE clusters, is delivered in a stacked follow-up PR (`feat/multi-cluster-az`).

### Changes

**API (additive, backwards-compatible):**
- `ProxmoxCluster.spec.availabilityZones` — new top-level field, a map keyed by `name` (`AvailabilityZoneSpec {name, nodes}`).
- `ProxmoxCluster.status.failureDomains` (`[]clusterv1.FailureDomain`) — reconciled from `spec.availabilityZones`, never set by the user.
- `ProxmoxMachine.status.failureDomain` (`string`, `MinLength=1`/`MaxLength=256`) — recorded from the zone of the node the machine is scheduled on.
- v1alpha1 conversion: `availabilityZones`, `failureDomains`, and `failureDomain` are restored across the hub→spoke→hub round-trip (they do not exist in v1alpha1).

**Controller:**
- `reconcileFailureDomains` syncs `status.failureDomains` from `spec.availabilityZones` (all zones advertised as control-plane-eligible).

**Scheduler:**
- When `Machine.spec.failureDomain` is set, candidate nodes are intersected with the zone's nodes.
- Round-robin balancing is performed per availability zone (via a `nodeToZone` map), so a multi-node zone is deprioritized as a group during rolling upgrades rather than per individual node.

**VM lifecycle:**
- `FindVM`/`updateVMLocation` set `ProxmoxMachine.status.failureDomain` from the scheduled node's zone.

**Docs and templates:**
- `docs/advanced-setups.md` documents the single-cluster availability-zone setup.
- The default and Cilium ClusterClasses expose an `availabilityZones` variable patched into the `ProxmoxClusterTemplate`.

### Relationship to #720

This PR covers the same CAPI contract surface as @emanuelebosetti's #720 — `status.failureDomains`, the machine failure-domain field, and `reconcileFailureDomains`. We are grateful to @emanuelebosetti for opening #717, for the working implementation, and for the research that established the contract mapping and the review feedback it collected. We reuse that foundation but take a different design direction on a few points:

| Aspect | #720 | This PR |
|---|---|---|
| API surface | Reuses `zoneConfig`, adding `nodes` and `controlPlane` to `ZoneConfigSpec` | Introduces a separate `spec.availabilityZones` (`AvailabilityZoneSpec`), keeping node grouping apart from `zoneConfig`'s IP/network config |
| Machine failure-domain field | `*string` on `ProxmoxMachineSpec` | `string` on `ProxmoxMachineStatus` with length validation, set by the controller from the scheduled node's zone |
| Control-plane eligibility | `ControlPlaneEligible` per zone | Not introduced; all zones are advertised as control-plane eligible |
| `FailureDomainNotReady` condition | Retryable condition with 30s requeue | Not required; the failure domain is read directly and recorded on status |
| Scheduler balancing | Per node | Per availability zone |
| Multi-DC (#370) | Explicitly out of scope, deferred to v1alpha3 | Delivered in a stacked follow-up PR via per-zone `credentialsRef` |

The net difference is that node grouping lives in a dedicated `availabilityZones` field, the machine failure domain is a status observation rather than a spec input, and balancing is zone-aware. Multi-cluster support is split into a follow-up to keep this PR within the project's review-size guidance.

### Backwards compatibility

Clusters without `availabilityZones` behave exactly as before. No feature gate, webhook change, or controller restart is required.

### Testing performed

- `make manifests generate` run; CRDs, deepcopy, and conversion outputs regenerated.
- `go build ./...` succeeds.
- `go test ./api/v1alpha1/... -run TestFuzzyConversion` passes (hub→spoke→hub round-trip for `ProxmoxCluster`, `ProxmoxMachine`, `ProxmoxClusterTemplate`, `ProxmoxMachineTemplate`, with `availabilityZones`/`failureDomains`/`failureDomain` normalized for fuzz stability).
- `go test ./internal/service/scheduler/...` passes, including `TestSelectNodeBalancesAcrossAvailabilityZones` (a two-node zone is deprioritized as a group so the second VM lands in the other zone) and the existing `TestSelectNode`/`TestScheduleVM` cases (unchanged behaviour when no zones are configured).
- `go test ./internal/service/vmservice/... ./internal/service/taskservice/... ./pkg/scope/...` passes.
- `api/v1alpha2` types test `TestSetInClusterIPPoolRefMultipleZones` passes.
- The `TestAPIs` envtest suites (`api/v1alpha1`, `api/v1alpha2`, `internal/controller`) were not executed in this environment because the kubebuilder envtest binaries are not installed; they should be run in CI.


vidarno:
_Disclosure: I did this review with Claude (Fable 5) as a pair. It pulled the branch, ran make lint verify test, and wrote the reproducer snippets below; I've checked each finding and the recommendations are mine. Happy to dig into any of them._

Thanks for splitting this out of #887 — the single-cluster failure-domain contract is the right first step, and the API shape (`status.failureDomain` on the machine, `availabilityZones` kept apart from `zoneConfig`) is a good foundation. I pulled the branch and ran it locally; a few things need fixing before it's mergeable, then some design points worth deciding now because #889 builds on them.

### Blocking

**1. CI `make verify` fails.** `make fix` rewrites `zoneForNode` in `internal/service/vmservice/find.go` to `slices.Contains`, leaving an uncommitted diff, so `verify` exits 1. `make lint` also fails on `vmscheduler_test.go:56` (`prealloc`). Please run `make lint verify test` before pushing (see AGENTS.md).

**2. `status.failureDomain` is never set on the normal path.** The only writer is `updateVMLocation` (`find.go:120`), which runs only in the VM-not-found/migration recovery branch. `createVM` isn't touched in this PR — the `Status.FailureDomain = scope.ZoneForNode(node)` line landed in #889 instead. Freshly provisioned machines will show no failure domain until #889 merges. That line belongs here.

### Should fix before merge

**3. Within-zone balancing regression.** Per-zone counting means every node in a zone gets the same `ScheduledVMs`, and `sortByReplicas.Less` has no tiebreak, so placement inside a zone degrades to memory-first. Reproducer with the existing helpers:

```go
azs := []infrav1.AvailabilityZoneSpec{{Name: "az-1", Nodes: []string{"pve1", "pve2"}}}
allowed := []string{"pve1", "pve2"}
locs := []infrav1.NodeLocation{{Node: "pve1"}, {Node: "pve1"}, {Node: "pve1"}}
mem := map[string]uint64{"pve1": miBytes(65536), "pve2": miBytes(65536)}
// with AZs: picks pve1 (already 3 VMs); with nodeToZone(nil, allowed): picks pve2
```

Any MD pinned to a multi-node zone loses node round-robin. Suggest sorting by `(zoneCount, nodeCount)` or adding a node-count tiebreak.

**4. Two different node→zone resolutions.** `zoneForNode` (find.go) is first-match; `nodeToZone` (scheduler) is last-match. A node listed in two AZs is counted for one zone and recorded in `status.failureDomain` as another. Please use one shared helper and add validation (webhook or CEL) that a node appears in at most one AZ — in #889 this ambiguity decides which Proxmox cluster receives the clone.

**5. No validation.** The docs say "every node listed in a zone must also be present in `allowedNodes`", but nothing enforces it?

**6. Verify placement against the contract.** After scheduling, if `Machine.spec.failureDomain` is set and `zoneForNode(target) != failureDomain`, return an error rather than recording a status that contradicts the spec. Cheap, and it would have caught 4.

### Minor

- `reconcileFailureDomains` only runs on the `reconcileNormal` success path, after IPAM; a cluster stuck on IPAM never publishes FDs. Consider running it before the early returns.
- Tests: there's none for `status.failureDomain` being written in `createVM` (finding 2).
- When #889 lands, the ClusterClass `availabilityZones` variable schema must grow `credentialsRef`; worth a TODO here so it isn't forgotten.

